PR #21, opened 2017-01-15
Added installation code for non-emscripten bin, lib and includes

nyue:


waldyrious:
I was about to suggest this! Just tested this PR branch locally and it works without issues.

CMakeLists.txt
```diff
@@ -526,4 +526,31 @@ else ()
                         "${GTEST_SOURCE_DIR}/googletest/include")
     target_link_libraries(draco_tests draco gtest)
   endif ()
+
+  # Headers installation
+  file ( GLOB_RECURSE draco_headers RELATIVE ${CMAKE_SOURCE_DIR} "*.h" )
```
waldyrious: maybe `CMAKE_CURRENT_SOURCE_DIR` instead?

tomfinegan: We use ${draco_root} to refer to the root of the Draco tree in the Draco cmake files.

CMakeLists.txt
```diff
@@ -526,4 +526,31 @@ else ()
...
+  foreach (filename ${draco_headers})
+    get_filename_component ( file_directory ${filename} DIRECTORY )
```
tomfinegan: style nit, please collapse unnecessary horizontal white space, this:
get_filename_component ( file_directory ${filename} DIRECTORY )

should be:
get_filename_component(file_directory ${filename} DIRECTORY)

CMakeLists.txt
```diff
@@ -526,4 +526,31 @@ else ()
...
+    install ( FILES
```
tomfinegan: nit: This can fit on one, at max two lines.

tomfinegan: ditto on collapsing horizontal whitespace, same for the following commands. The foreach and endforeach are the only commands in here that should have a space between the command and the open parenthesis given the style throughout the rest of the cmake file.

CMakeLists.txt
```diff
@@ -526,4 +526,31 @@ else ()
...
+      ${filename}
+      DESTINATION
+      ${CMAKE_INSTALL_PREFIX}/include/${file_directory}
```
tomfinegan: This should be include/draco/${file_directory}; I don't think we want the top level of the draco includes in /usr/local/include; we probably want them in /usr/local/include/draco.

tomfinegan: Closing PR due to no response on comments.

PR #37, opened 2017-01-18
Fix setting stream offset in mesh read

rtoepfer:
Fix stream read based from original offset which matches vector size.


ondys:
Thanks for the pull request. We have already fixed the reported issue and the fix has been pushed to our public repo.

Just one note about your pull request. I believe the following line would not be valid:
`is.seekg(0, is_size);`

when `seekg` is used with two arguments, the second argument needs to be one of the predefined constants:
`ios_base::beg` , `ios_base::cur`, or `ios_base::end`

PR #45, opened 2017-01-24
Javascript loader: load texture coordinates (uvs)

customlogic:
The DRACOLoader.js file was ignoring texture coordinates (uvs) when loading objects. This pull request correctly loads the texture coordinates and applies them to the BufferGeometry.

### DIFF
--- a/javascript/example/DRACOLoader.js
+++ b/javascript/example/DRACOLoader.js
@@ -75,7 +75,7 @@ THREE.DRACOLoader.prototype = {
         /*
          * Example on how to retrieve mesh and attributes.
          */
-        let numFaces, numPoints, numVertexCoordinates, numAttributes;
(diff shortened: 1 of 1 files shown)


javascript/example/DRACOLoader.js
```diff
@@ -161,12 +174,24 @@ THREE.DRACOLoader.prototype = {
                   norAttributeData.GetValue(i + 2));
             }
         }
+
+        for (let i = 0; i < numTextureCoordinates; i += 2) {
+            // Add texture coordinates.
+            if (texCoordAttId != -1) {
```
ondys: This line (180) should be before the for loop ( we don't want to loop through all the numTextureCoordinates when texCoordAttId is not used).

customlogic: Ah, yeah. I've fixed and pushed commit

PR #73, opened 2017-03-06
fixed threejs example vertex color normal

mackuntu:
Current example loader will not display vertex colors correctly, since color attribute data need to be normalized as a fraction of 255.


mackuntu:
I signed it!

mackuntu:
I really signed it

laurielabs:
confirmed

PR #74, opened 2017-03-06
fixed threejs example vertex color normal

laurielabs:
Current example loader will not display vertex colors correctly, since color attribute data need to be normalized as a fraction of 255.


javascript/example/DRACOLoader.js
```diff
@@ -177,10 +177,12 @@ THREE.DRACOLoader.prototype = {
             geometryBuffer.vertices[i + 1] = posAttributeData.GetValue(i + 1);
             geometryBuffer.vertices[i + 2] = posAttributeData.GetValue(i + 2);
             // Add color.
+            // 03/05/2017, mackuntu@: ThreeJS vertex colors need to be
```
FrankGalligan: Can you change this to not have the date and username. We don't usually put usernames in the comments unless it is a TODO.

If you want, you can add your username to the AUTHORS file.

laurielabs: done

PR #78, opened 2017-03-16
Simplify steps required to run example.

donmccurdy:
I'd like to suggest simplifying the steps required to get an example running for newer users.

The proposed change would link to the latest stable version of three.js (r84) from a CDN, and use a relative link to the existing `draco_decoder.js` file.

### DIFF
--- a/javascript/example/webgl_loader_draco.html
+++ b/javascript/example/webgl_loader_draco.html
@@ -41,8 +41,8 @@
         </div>
         <pre id="fileDisplayArea"><pre>
     </div>
-    <script src="three.js"></script>
-    <script src="draco_decoder.js"></script>
+    <script src="https://cdn.rawgit.com/mrdoob/three.js/r84/build/three.min.js"></script>
+    <script src="../draco_decoder.js"></script>
     <script src="DRACOLoader.js"></script>
     <script>
       'use strict';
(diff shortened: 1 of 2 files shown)


javascript/example/webgl_loader_draco.html
```diff
@@ -41,8 +41,8 @@
         </div>
         <pre id="fileDisplayArea"><pre>
     </div>
-    <script src="three.js"></script>
-    <script src="draco_decoder.js"></script>
+    <script src="https://cdn.rawgit.com/mrdoob/three.js/r84/build/three.min.js"></script>
+    <script src="../draco_decoder.js"></script>
```
ondys: this relative path is inaccessible to the server when it is running from the javascript/example/ folder (at least to the http server that we suggest in our documentation: `python -m SimpleHTTPServer`

donmccurdy: Yes, my changes to the instructions (attempt to) specify that the local server should be run from the root of the cloned folder, instead. Is this compatible with your own workflows?

ondys: We discussed it and it should be ok. It may be better to make it more explicit that the server needs to be run from either the root location or the javascript/ folder (update the new step (2)).

donmccurdy: Sounds good, is this clearer?

ondys:
Sounds good. Thanks for the PR

Issue #83, opened 2017-03-18
OSX build is broken


asluk:
OSX builds get the following compiler error:

raco/core/vector_d_test.cc:90:27: error: use of undeclared identifier 'sqrt'
  const float magnitude = sqrt(v.SquaredNorm());

I have a fix.

ondys:
Thanks for the fix.

PR #84, opened 2017-03-18
Fix build for sqrt usage in vector.

asluk:
Adding cmath include to fix OSX build (issue #83).

### DIFF
--- a/core/vector_d.h
+++ b/core/vector_d.h
@@ -18,6 +18,7 @@
 #include <inttypes.h>
 #include <algorithm>
 #include <array>
+#include <cmath>
 
 #include "core/macros.h"
 


ondys:
Thanks for catching and fixing this

jwakely:
Since `sqrt` is called unqualified you need to include `<math.h>` not `<cmath>`. Either that or add a using-declaration. There's no guarantee that `<cmath>` declares `sqrt` in the global namespace.

ondys:
@jwakely In our internal version, we have already changed this to std::sqrt which I believe requires <cmath>

jwakely:
@ondys right, if you use `std::sqrt` you should include `<cmath>` and if you use `::sqrt` you should include `<math.h>` so sounds like it's correct now. Jolly good!

PR #85, opened 2017-03-26
Expose .isVersionSupported() method on THREE.DRACOLoader.

donmccurdy:
I wanted to avoid making another instance of DracoModule just to check a version, so this PR also exposes a singleton decoder with a static factory method.

(good use case for adding `?w=1` to the URL to ignore whitespace in diffs)

### DIFF
--- a/javascript/example/DRACOLoader.js
+++ b/javascript/example/DRACOLoader.js
@@ -265,7 +265,7 @@ THREE.DRACOLoader.prototype = {
     },
 
     isVersionSupported: function(version) {
-      return this.decoder.isVersionSupported(version);
+        return this.decoder.isVersionSupported(version);
     }
 };
 
@@ -275,16 +275,16 @@ THREE.DRACOLoader.prototype = {
  */
 THREE.DRACOLoader.getDecoder = (function() {
 
-  let decoder;
+    let decoder;
 
-  if (typeof DracoModule === 'function') {
-    decoder = DracoModule();
-  } else {
-    console.error('THREE.DRACOLoader: DracoModule not found.');
-  }
+    if (typeof DracoModule === 'function') {
+      decoder = DracoModule();
+    } else {
+      console.error('THREE.DRACOLoader: DracoModule not found.');
+    }
 
-  return function() {
-    return decoder;
-  };
+    return function() {
+        return decoder;
+    };
 
 })();
--- a/javascript/example/DRACOLoader.js
+++ b/javascript/example/DRACOLoader.js
@@ -15,7 +15,6 @@
 'use strict';
 
 THREE.DRACOLoader = function(manager) {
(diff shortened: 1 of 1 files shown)


javascript/example/DRACOLoader.js
```diff
@@ -15,6 +15,7 @@
 'use strict';
 
 THREE.DRACOLoader = function(manager) {
+    this.decoder = THREE.DRACOLoader.getDecoder();
```
ondys: is it necessary to store the decoder as a member variable? Isn't `THREE.DRACOLoader.getDecoder()` sufficient? (i.e. replacing instances of `this.decoder;` with `this.getDecoder();`)

donmccurdy: Instances of `this.decoder` would need to be replaced with the fully explicit `THREE.DRACOLoader.getDecoder()`, as JS only searches up the prototype chain, but you're correct that the member variable isn't necessary. I'm happy to do either, but this felt a bit cleaner to me.

ondys: The only concern I have about having it as a member variable is that the `DracoModule` would be automatically instantiated when `THREE.DRACOLoader` is created even if there is no geometry decoded later on. I'm not sure if that's really a problem in practice though. I think it's ok for now, we can change it later if there are some problems with this approach. I'll merge it later today. Thanks!

donmccurdy: Oh, some clarification on the immediately invoked functions `(function(){})()` in JavaScript: both with and without this PR, the `DracoModule` is instantiated when the script loads — even before the `new THREE.DRACOLoader()` is created.

Happy to fix that while we're here, though. How does this look? I also switch from logging an error to throwing an error, since it's now happening at invocation and would throw an (less clear) error later regardless.

ondys:
Thanks for the PR!

javascript/example/DRACOLoader.js
ondys: Thanks for the clarification. Looks good to me.

We should still merge it to the official THREE.js repo. Eventually it would be good idea to only maintain the Draco loader there, but let's keep it duplicate for now.

donmccurdy: Thanks, I'll get it updated in the three.js repo as well. 

I agree that only maintaining THREE.DRACOLoader there would be probably better; do you have specific tasks ought to happen first? E.g. does the API need to be more finalized first? Can start another issue to discuss, too.

ondys: We currently have some internal changes that we would like to push to our git first to get them tested before they are integrated to the official THREE.js repo, but in general I think we should be able to fully migrate the loader to THREE.js repo pretty soon.

donmccurdy: Sounds good!

PR #106, opened 2017-05-03
Update parser_utils.cc

awmcc90:
When using isspace(c), c should be an unsigned char given __chvalidator() requires a valid ascii value ( c >= -1 && c <= 255) and throws an error in the case of a signed char value between -128 and -2.


awmcc90:
I signed it

ondys:
Thanks for the PR!

PR #114, opened 2017-05-25
Fix build on Android with gnustl

KindDragon:


KindDragon:
> gnustl doesn't define std::log2 in ?

Yes, https://github.com/android-ndk/ndk/issues/82

> are you linking to the c++11 version of the library?

Yes with `-std=c++11`

ondys:
Thanks for the PR. We will change it for now. Hopefully gnustl will add std::log2 soon.

mekya:
btw, for the release 1.2.0 
options.h uses `std::to_string` and `std::strtof` functions which are not in gnustl

PR #135, opened 2017-07-23
Update bitstream spec

FrankGalligan:
This PR is associated with #134.

### DIFF
--- a/docs/spec/attributes.decoder.md
+++ b/docs/spec/attributes.decoder.md
@@ -5,7 +5,7 @@
 
 ~~~~~
 DecodeAttributesDecoderData(buffer) {
-  num_attributes                                                        I32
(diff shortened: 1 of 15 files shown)


docs/spec/attributes.decoder.md
```diff
@@ -5,7 +5,7 @@
 
 ~~~~~
 DecodeAttributesDecoderData(buffer) {
-  num_attributes                                                        I32
+  num_attributes                                                        varUI32
```
ondys: do we have "varUI32" documented anywhere? (for example that it is decoded with  need to decode it with DecodeVarint<UI32>() function)

louquillio: @ondys, I'm going to merge this. Could you open an issue to track this concern? Thanks.

PR #157, opened 2017-08-01
Simplify Node.js usage example.

donmccurdy:
Couple small changes here:

- Replace `nodejs` with more common `node` executable name.
- Add GitHub URL to npm package, so cross-link will appear on [package description](https://www.npmjs.com/package/draco3d).
- Simplify example script a bit, reusing code shared among attributes, and fixing a couple let/const issues.

### DIFF
--- a/javascript/npm/draco3d/draco_nodejs_example.js
+++ b/javascript/npm/draco3d/draco_nodejs_example.js
@@ -67,9 +67,10 @@ function encodeMeshToFile(mesh, decoder) {
   const numIndices = numFaces * 3;
   const numPoints = mesh.num_points();
   const numVertexCoord = numPoints * 3;
+  const indices = new Uint32Array(numIndices);
+
   console.log("Number of faces " + numFaces);
   console.log("Number of vertices " + numPoints);
-  const indices = new Uint32Array(numIndices);
 
   // Add Faces to mesh
   const ia = new decoderModule.DracoInt32Array();
@@ -83,9 +84,10 @@ function encodeMeshToFile(mesh, decoder) {
   decoderModule.destroy(ia);
   meshBuilder.AddFacesToMesh(newMesh, numFaces, indices);
 
-  const attrs = ["POSITION", "NORMAL", "TEX_COORD", "COLOR"];
+  const attrs = {POSITION: 3, NORMAL: 3, COLOR: 3, TEX_COORD: 2};
 
-  attrs.forEach((attr) => {
+  Object.keys(attrs).forEach((attr) => {
+    const stride = attrs[attr];
     const decoderAttr = decoderModule[attr];
     const encoderAttr = decoderModule[attr];
     const attrId = decoder.GetAttributeId(mesh, decoderAttr);
@@ -103,15 +105,15 @@ function encodeMeshToFile(mesh, decoder) {
     assert(numVertexCoord === attributeData.size(), 'Wrong attribute size.');
 
     const attributeDataArray = new Float32Array(numVertexCoord);
-    for (let i = 0; i < numVertexCoord; i += 3) {
-      attributeDataArray[i] = attributeData.GetValue(i);
-      attributeDataArray[i + 1] = attributeData.GetValue(i + 1);
-      attributeDataArray[i + 2] = attributeData.GetValue(i + 2);
+    for (let i = 0; i < numVertexCoord; i += stride) {
+      for (let j = 0; j < stride; ++j) {
+        attributeDataArray[i + j] = attributeData.GetValue(i + j);
+      }
     }
 
     decoderModule.destroy(attributeData);
-    meshBuilder.AddFloatAttributeToMesh(newMesh, encoderAttr, numPoints, 3,
-        attributeDataArray);
+    meshBuilder.AddFloatAttributeToMesh(newMesh, encoderAttr, numPoints,
+        stride, attributeDataArray);
   });
 
   let encodedData = new encoderModule.DracoInt8Array();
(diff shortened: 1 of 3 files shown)


javascript/npm/draco3d/draco_nodejs_example.js
```diff
@@ -82,81 +83,36 @@ function encodeMeshToFile(mesh, decoder) {
...
+    decoder.GetAttributeFloatForAllPoints(mesh, attribute, attributeData);
+
+    assert(numVertexCoord === attributeData.size(), 'Wrong attribute size.');
+
+    const attributeDataArray = new Float32Array(numVertexCoord);
     for (let i = 0; i < numVertexCoord; i += 3) {
-      colors[i] = colAttributeData.GetValue(i);
-      colors[i + 1] = colAttributeData.GetValue(i + 1);
-      colors[i + 2] = colAttributeData.GetValue(i + 2);
+      attributeDataArray[i] = attributeData.GetValue(i);
```
fanzhanggoogle: I believe I made a mistake here that texture coordinates should not have three vertex data. In which case we can not use 3 for size of per attribute data. It's my bad that I didn't test a model with texture coordinates. This PR LGTM if you could fix this? I think a simple way will be just check if decoderAttr === decoderModule.TEX_COORD then change the size of per vertex data?

donmccurdy: Good catch! Fixed.

javascript/npm/draco3d/draco_nodejs_example.js
```diff
@@ -82,81 +84,37 @@ function encodeMeshToFile(mesh, decoder) {
...
-      colors[i] = colAttributeData.GetValue(i);
-      colors[i + 1] = colAttributeData.GetValue(i + 1);
-      colors[i + 2] = colAttributeData.GetValue(i + 2);
+    console.log("Adding %s attribute", attr);
+
+    const attribute = decoder.GetAttribute(mesh, attrId);
+    const attributeData = new decoderModule.DracoFloat32Array();
+    decoder.GetAttributeFloatForAllPoints(mesh, attribute, attributeData);
+
+    assert(numVertexCoord === attributeData.size(), 'Wrong attribute size.');
```
ondys: Is this going to work for texture coordinates? 
`numVertexCoords` is set to be `numPoints * 3`  , but texture coordinates have only `numPoints * 2` values so it would probably fail .. maybe 
assert(numPoints * stride === attributeData.size())

javascript/npm/draco3d/draco_nodejs_example.js
```diff
@@ -82,81 +84,37 @@ function encodeMeshToFile(mesh, decoder) {
...
+
+    const attributeDataArray = new Float32Array(numVertexCoord);
+    for (let i = 0; i < numVertexCoord; i += stride) {
```
ondys: `numVertexCoord` should be probably `numPoints * stride` .. 

this can be also only one loop where you increment '++i'

javascript/npm/draco3d/draco_nodejs_example.js
donmccurdy: Fixed.

javascript/npm/draco3d/draco_nodejs_example.js
donmccurdy: First issue is fixed.

I don't see what you mean on the second  — the inner loop uses `j`.

javascript/npm/draco3d/draco_nodejs_example.js
```diff
@@ -82,81 +84,37 @@ function encodeMeshToFile(mesh, decoder) {
...
-    decoderModule.destroy(norAttributeData);
-    meshBuilder.AddFloatAttributeToMesh(newMesh, encoderModule.NORMAL,
-        numPoints, 3, normals);
-  }
+  const attrs = {POSITION: 3, NORMAL: 3, COLOR: 3, TEX_COORD: 2};
+
+  Object.keys(attrs).forEach((attr) => {
+    const stride = attrs[attr];
+    const decoderAttr = decoderModule[attr];
+    const encoderAttr = decoderModule[attr];
```
ondys: I'm not sure if it really matters but for `encoderAttr` , this should be probably taken from `encoderModule[attr]`

javascript/npm/draco3d/draco_nodejs_example.js
```diff
@@ -82,81 +84,37 @@ function encodeMeshToFile(mesh, decoder) {
...
-    decoder.GetAttributeFloatForAllPoints(mesh, texAttribute,
-        texAttributeData);
-    assertEquals(numVertexCoord, texAttributeData.size());
-    const texcoords = new Float32Array(numVertexCoord);
-    for (let i = 0; i < numVertexCoord; i += 3) {
-      texcoords[i] = texAttributeData.GetValue(i);
-      texcoords[i + 1] = texAttributeData.GetValue(i + 1);
-      texcoords[i + 2] = texAttributeData.GetValue(i + 2);
+    if (attrId < 0) {
+      return;
```
ondys: I don't think we should return if we don't find the attribute .. we can just ignore them if they are not there (probably except for the position attribute)

javascript/npm/draco3d/draco_nodejs_example.js
ondys: I meant that we don't need the inner loop .. I'm ok keeping it if you think it helps readability, but we can just iterate over every single element in the first loop instead of using `i += stride`

donmccurdy: Oh, yes that's better. 🙂

javascript/npm/draco3d/draco_nodejs_example.js
donmccurdy: Typo; good point.

Perhaps for later, but if they really are the same then maybe only one should be used in the example... it is a bit confusing having both.

javascript/npm/draco3d/draco_nodejs_example.js
donmccurdy: This is in the `forEach` callback, so returning is equivalent to `continue` in a loop here. Missing attributes are ignored.

ondys:
Thanks for the PR!

Issue #171, opened 2017-08-05
Decoder derefereces empty unique ptr when decoding drc file that was created with -qp 0


pps83:
I created a drc file with -qp 0 (see attached inside zip at the bottom).

When I try to decode it to .obj draco_decoder takes long time and then crashes. It appears that it dereferences portable_attribute_ ptr without checking if it's not empty:

![image](https://user-images.githubusercontent.com/1614246/28999097-093a6d24-79f0-11e7-9500-7f1413612703.png)

Then only place that may set `portable_attribute_`  seem to be `SequentialAttributeDecoder::SetPortableAttribute` and before that code that crashes this function was never called.

more detailed backtrace:
```
>	draco_decoder.exe!draco::PointAttribute::is_mapping_identity() Line 55	C++
 	draco_decoder.exe!draco::SequentialAttributeDecoder::GetPortableAttribute() Line 61	C++
 	draco_decoder.exe!draco::SequentialAttributeDecodersController::GetPortableAttribute(int point_attribute_id) Line 43	C++
 	draco_decoder.exe!draco::PointCloudDecoder::GetPortableAttribute(int parent_att_id) Line 157	C++
 	draco_decoder.exe!draco::SequentialAttributeDecoder::InitPredictionScheme(draco::PredictionSchemeInterface * ps) Line 83	C++
 	draco_decoder.exe!draco::SequentialIntegerAttributeDecoder::DecodeValues(const std::vector<draco::IndexType<int,draco::PointIndex_tag_type_>,std::allocator<draco::IndexType<int,draco::PointIndex_tag_type_> > > & point_ids, draco::DecoderBuffer * in_buffer) Line 55	C++
 	draco_decoder.exe!draco::SequentialAttributeDecoder::DecodePortableAttribute(const std::vector<draco::IndexType<int,draco::PointIndex_tag_type_>,std::allocator<draco::IndexType<int,draco::PointIndex_tag_type_> > > & point_ids, draco::DecoderBuffer * in_buffer) Line 41	C++
 	draco_decoder.exe!draco::SequentialAttributeDecodersController::DecodePortableAttributes(draco::DecoderBuffer * in_buffer) Line 66	C++
 	draco_decoder.exe!draco::AttributesDecoder::DecodeAttributes(draco::DecoderBuffer * in_buffer) Line 55	C++
 	draco_decoder.exe!draco::SequentialAttributeDecodersController::DecodeAttributes(draco::DecoderBuffer * buffer) Line 60	C++
[...]

Issue #108, opened 2017-05-09
ReadMeshFromFile assumes file name has at least 4 characters


jwakely:
$ ./draco_encoder -i ab
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::substr: __pos (which is 18446744073709551614) > this->size() (which is 2)
Aborted (core dumped)

````
std::unique_ptr<Mesh> ReadMeshFromFile(const std::string &file_name) {
  std::unique_ptr<Mesh> mesh(new Mesh());
  // Analyze file extension.
  const std::string extension =
      parser::ToLower(file_name.substr(file_name.size() - 4));
````

**Comment by pps83** [comment]
Because they do this:
```
  const std::string extension =
      parser::ToLower(file_name.substr(file_name.size() - 4));
```
which is complete WTF. It's been months and nobody even cared to check this issue report I guess.

PR #173, opened 2017-08-07
fix for #108 and workaround to avoid crash in #171

pps83:
there is definitely a bug somewhere that led to #171, perhaps shouldn't be included if actual bug is fixed.


pps83:
I signed it!

PR #179, opened 2017-08-11
npm: fixed npm example

sjfricke:
Allows draco3d to be called from same directory or npm package

If you follow the [instructions on NPM readme](https://github.com/google/draco/blob/master/javascript/npm/draco3d/README.md#npm-package) you will find you need to edit the file because `./draco3d` is not a file when you copy it out of the `node_module` file with `cp node_modules/draco3d/draco_nodejs_example.js .`

This seems like the simpler solution than adding an instruction to open the file in a text editor and fix it

### DIFF
--- a/javascript/npm/draco3d/draco_nodejs_example.js
+++ b/javascript/npm/draco3d/draco_nodejs_example.js
@@ -16,7 +16,7 @@
 
 const fs = require('fs');
 const assert = require('assert');
-const draco3d = require('./draco3d');
+const draco3d = fs.existsSync('./draco3d') ? require('./draco3d') : require('draco3d');
 const decoderModule = draco3d.createDecoderModule({});
 const encoderModule = draco3d.createEncoderModule({});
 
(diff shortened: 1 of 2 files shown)


sjfricke:
Also fixed README path to source and added a more GitHub friendly way since whats the point of giving a path on a markdown page viewed online without having link in already.

javascript/npm/draco3d/draco_nodejs_example.js
```diff
@@ -16,7 +16,7 @@
 
 const fs = require('fs');
 const assert = require('assert');
-const draco3d = require('./draco3d');
+const draco3d = fs.existsSync('./draco3d') ? require('./draco3d') : require('draco3d');
```
fanzhanggoogle: Thanks for the fix. You are right. But I think we don't need to check existsSync. Just the following should be fine for both cases:
~~~
const draco3d = require('draco3d');
~~~

sjfricke: your right, I guess I never checked that `draco3d` would check locally as well

sjfricke:
Updated with way better and simpler line

fanzhanggoogle:
Thanks!

PR #182, opened 2017-08-17
Added handling for all enum in draco::DataType

nyue:
Added support for 64bit integer, 64bit float and boolean in
ConvertValue
Reordered the case label to match those in the enum
for easier maintenance in the long run

### DIFF
--- a/src/draco/attributes/geometry_attribute.h
+++ b/src/draco/attributes/geometry_attribute.h
@@ -112,40 +112,40 @@ class GeometryAttribute {
     if (out_val == nullptr)
       return false;
     switch (data_type_) {
-	  case DT_INT8:
+      case DT_INT8:
         return ConvertTypedValue<int8_t, OutT, out_att_components_t>(att_id,
                                                                      out_val);
-	  case DT_UINT8:
-		  return ConvertTypedValue<uint8_t, OutT, out_att_components_t>(att_id,
-                                                                        out_val);
+      case DT_UINT8:
+        return ConvertTypedValue<uint8_t, OutT, out_att_components_t>(att_id,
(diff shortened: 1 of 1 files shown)


src/draco/attributes/geometry_attribute.h
```diff
@@ -112,28 +112,40 @@ class GeometryAttribute {
...
       return false;
     switch (data_type_) {
-      case DT_FLOAT32:
-        return ConvertTypedValue<float, OutT, out_att_components_t>(att_id,
-                                                                    out_val);
-      case DT_UINT8:
-        return ConvertTypedValue<uint8_t, OutT, out_att_components_t>(att_id,
-                                                                      out_val);
-      case DT_INT8:
+	  case DT_INT8:
```
ondys: Please, use the indent of the original code (no tabs, two space indent relative to the previous block)

src/draco/attributes/geometry_attribute.h
```diff
@@ -112,28 +112,40 @@ class GeometryAttribute {
...
         return ConvertTypedValue<int8_t, OutT, out_att_components_t>(att_id,
                                                                      out_val);
-      case DT_UINT16:
-        return ConvertTypedValue<uint16_t, OutT, out_att_components_t>(att_id,
-                                                                       out_val);
+	  case DT_UINT8:
+		  return ConvertTypedValue<uint8_t, OutT, out_att_components_t>(att_id,
```
ondys: same  for the `return`,  every new block should start two spaces from the previous one

nyue:
I have updated the changes from tab to spaces

ondys:
Thanks for the PR

PR #257, opened 2017-10-24
Spec: Fix Tagged Symbols constant

FrankGalligan:
- Addresses a comment in KhronosGroup/glTF#1114


docs/spec/rans.decoding.md
```diff
@@ -26,7 +26,7 @@ void DecodeTaggedSymbols(num_values, num_components, out_values) {
   encoded_data                                                                        UI8[size]
   RansInitDecoder(ans_, &encoded_data[0], size, l_rans_base);
```
lexaknyazev: Should `l_rans_base` here also be replaced with constant value?

FrankGalligan: yes, will update now.

PR #301, opened 2017-11-19
Optional Compiler Optimization

eddiz:
Enable all Compile Optimization by setting according compiler flags. (Visual Studio tested, gcc and other should work properly, but untested)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,17 @@ if (ENABLE_GOMA)
   set_compiler_launcher(ENABLE_GOMA gomacc)
 endif ()
 
+if(ENABLE_EXTRA_SPEED)
+	if(MSVC)
+		#enable parallel builds
+		add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/Qpar>")
+		#maximum optimization in Release mode
+		add_compile_options("$<$<AND:$<CXX_COMPILER_ID:MSVC>,$<CONFIG:RELEASE>>:/Ox>")
+	else()
+		add_compiler_flag_if_supported("-O3")
+	endif()
+endif()
+
 # Generate a version file containing repository info.
 include(FindGit)
 find_package(Git)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,14 +102,14 @@ if (ENABLE_GOMA)
 endif ()
 
 if(ENABLE_EXTRA_SPEED)
-	if(MSVC)
-		#enable parallel builds
-		add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/Qpar>")
-		#maximum optimization in Release mode
-		add_compile_options("$<$<AND:$<CXX_COMPILER_ID:MSVC>,$<CONFIG:RELEASE>>:/Ox>")
-	else()
-		add_compiler_flag_if_supported("-O3")
-	endif()
+  if(MSVC)
+    #maximum optimization in Release mode
+    add_compile_options("$<$<CONFIG:RELEASE>:/Ox>")
+	#add_compile_options("$<$<CONFIG:MinSizeRel>:/Ox>") #uncomment if consistent compiler settings are more important then binary size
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -101,6 +101,17 @@ if (ENABLE_GOMA)
   set_compiler_launcher(ENABLE_GOMA gomacc)
 endif ()
 
+if(ENABLE_EXTRA_SPEED)
+	if(MSVC)
```
tomfinegan: Indent is 2 spaces in the Draco CMake files.

CMakeLists.txt
```diff
@@ -101,6 +101,17 @@ if (ENABLE_GOMA)
...
+		#enable parallel builds
```
tomfinegan: Please fix this comment: it doesn't explain what /Qpar is actually doing. Microsoft documentation states the following:

Enables the Auto-Parallelizer feature of the compiler to automatically parallelize loops in your code.

CMakeLists.txt
```diff
@@ -101,6 +101,17 @@ if (ENABLE_GOMA)
...
+		add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/Qpar>")
```
tomfinegan: Drop the extra generator expression. You're already within a block wrapper w/if(MSVC).

CMakeLists.txt
```diff
@@ -101,6 +101,17 @@ if (ENABLE_GOMA)
...
+		#maximum optimization in Release mode
+		add_compile_options("$<$<AND:$<CXX_COMPILER_ID:MSVC>,$<CONFIG:RELEASE>>:/Ox>")
```
tomfinegan: Same w/the extra MSVC check here. Please remove.

Also, I would prefer an if instead of a generator exp here. Something like:

if (CMAKE_BUILD_TYPE MATCHES "^Rel")
  add_compile_options(...)
endif ()

That way the flag gets added to Release and RelWithDebInfo while excluding MinSizeRel.

CMakeLists.txt
eddiz: sorry my fault, this line should be removed. Visual Studio's comment was irritating in this sense, stating: "Enable Parallel Code Generation"

CMakeLists.txt
eddiz: Got your point. However, please consider that the progress of compiling a CMake project using Visual Studio is different to CMake + gcc. In case of Visual Studio CMake builts the project once and CMake is not touched anymore afterward (unless you modified CMake files). Switching between Debug, Release etc. is happening in a way that CMake is not involved in this process at all. 

That flag should not be added for MinSizeRel Configurations, because /Ox increases the file size of the binary. In contrast /O1 optimizes for the size of the binary. Current CMake configuration handles this case already correct.
I am also not sure whether it is worth combining /Ox with RelWithDebInfo. The trade of here is higher compilation times vs. consistency with Release build.

CMakeLists.txt
eddiz: remove this line as well, se comment above

CMakeLists.txt
tomfinegan: > Got your point. However, please consider that the progress of compiling a CMake project using Visual Studio is different to CMake + gcc.

Err, yeah, sorry-- I've managed to forget that CMAKE_BUILD_TYPE is kind of useless w/MSVC because of the nature of the MSVC generator. Probably have to use the generator expression here.

> That flag should not be added for MinSizeRel Configurations

MinSizeRel will not match ^Rel; that's why the '^' was there. Moot point though since you probably have to ignore my suggestion and continue with generator use.

Please fix the indentation and comment, and then we can look into getting this merged. 

Also, have you accepted the [Google CLA](https://cla.developers.google.com/clas)? You'll need to do that before I can merge this.

eddiz:
@tomfinegan: build script is updated according to the discussion. Commented blocks are also tested - feel free to uncomment

Issue #306, opened 2017-11-24
Compilation errors when building for Android ArmV8 64 bit / Clang / LibC++


jherico:
Building against commit 441d5e05f74f405c309f6099df1fd8b71d90cc43 (current HEAD/1.2.2)

After configuring the application with the command 
```
$HOME/Android/SDK/cmake/3.6.4111459/bin/cmake .. -DCMAKE_INSTALL_PREFIX=$HOME/Android/build/nvtt -DANDROID_ABI=arm64-v8a -DANDROID_PLATFORM=android-24 -DCMAKE_BUILD_TYPE=Release -DANDROID_NDK=$HOME/Android/SDK/ndk-bundle -DCMAKE_TOOLCHAIN_FILE=$HOME/Android/SDK/ndk-bundle/build/cmake/android.toolchain.cmake -DCMAKE_MAKE_PROGRAM=$HOME/Android/SDK/cmake/3.6.4111459/bin/ninja -G"Android Gradle - Ninja" -DANDROID_PLATFORM=android-24 -DANDROID_TOOLCHAIN=clang -DANDROID_STL=c++_shared
```

I encountered a number of compile errors.  

In two locations (obj_decoder.cc and ply_decoder.cc) the code attempts the following assignment `file_size = file.tellg() - file_size;` where `file_size` is an `int64_t`.  This fails on the Clang builds unless you cast the result of `file.tellg()` to a `int64_t`.  Alternatively, declaring `file_size` as `auto` when initializing it from `file.tellg()` also solves the problem.

In `void Options::SetVector` in options.h, there is the following block

```
#ifdef ANDROID
    out += to_string(vec[i]);
#else
    out += std::to_string(vec[i]);
#endif
```

This is likely an attempt to work around some failing of one of the STL implementations in the NDK, but fails to compile with Clang / LibC++.  Locally I have eliminated the block, but presumably a more sophisticated `#if` statement would be required to allow compilation on multiple platforms.

The CMake function `add_cxx_preproc_definition` doesn't appear to function properly.  I had compile errors in decode.cc due to missing headers, which turned out to be due to `DRACO_POINT_CLOUD_COMPRESSION_SUPPORTED` not being defined.  I modified the add_cxx_preproc_definition function to replace the contents with simply `add_definitions(-D${preproc_def})` but this would apply to both C++ and C code, so an alternative approach may be required.

jherico:
[...]

PR #307, opened 2017-11-24
Fixing Android ArmV8/Clang/LibC++ build

jherico:
Changes required to build for Android ARMv8 with Clang and LibC++,  fixes for #306.  If I can get feedback on more platform neutral ways to fix these issues I will apply and update the PR.


jherico:
Signed CLA

Issue #312, opened 2017-12-04
Possible memory leak in DRACOLoader.js


simena86:
I'm using DRACOLoader.js to load draco files (files holds meshes encoded from .obj files):

``` javascript
const dracoLoader = new THREE.DRACOLoader('<url-to-gcs-bucket-with warm decoder>');
dracoLoader.setVerbosity(1);
dracoLoader.setCrossOrigin('anonymous');

...
let mesh = null;
let scene = null;

// initialize three.js scene
...

function loadMesh() {
    dracoLoader.load('<url to drc file>', function(bufferGeometry) {
        const material = new THREE.MeshPhongMaterial({
                    color: 0x996633,
                    specular: 0x050505,
                    shininess: 100
          });
        
          mesh = new THREE.Mesh(bufferGeometry, material);
          scene.add(mesh)
    })
}

function closeViewer() {
    scene.remove(mesh);
    mesh.geometry.dispose();
}

```

After having loaded the files, and closing my viewer I never seem to be able to release the memory used to load the files.

I've tried to NOT add the decoded geometry  to my THREE.js scene, just to try to rule out that the problem is with THREE.js. Memory is still not released. 
I tried to `dispose`  geometry and destroy all the draco objects DRACOLoader.js' `convertDracoGeometryTo3JS` function:

``` javascript
        convertDracoGeometryTo3JS: function(dracoDecoder, decoder, geometryType,
                                            buffer) {
            if (this.getAttributeOptions('position').skipDequantization === true) {
               decoder.SkipAttributeTransform(dracoDecoder.POSITION);
            }
            .....
            
            // try to destroy the objects
            dracoDecoder.destroy(posTransform);
            dracoDecoder.destroy(decoder);
            dracoDecoder.destroy(dracoGeometry);
            geometry.dispose();
            return null;
```

so my thought is that with the above code, the memory should've been released.  But from taking a memory snapshot with chrome, I get that more then 1GB of memory is still allocated:

![image](https://user-images.githubusercontent.com/1786123/33549253-319e3cd8-d8ea-11e7-9c0d-bde1814572b5.png)

The draco file is aprox: 6MB with 3295950 verts and 3024128 faces
I am using three.js v0.88.0

ondys:
Was the `dracoLoader` still allocated at the point you were checking the memory?

As long as the Draco module inside the loader is allocated, it may still hold to the internal memory that was used during decoding

simena86:
It probably was. I now tried to do the following to release memory:

[...]

PR #315, opened 2017-12-16
DRACOLoader: Simplify decoder source loading.

donmccurdy:
Simplifies loading of decoder source files:

1. Remove `setTimeout()` polling for the decoder module.
2. Use static `.setDecoderPath()` and `.setDecoderConfig()` methods to have a more standard JS API. I think this makes the singleton decoder lifecycle feel more natural.
3. Use Promises for internal methods, while keeping the same callback-based public API, which just simplifies the code a bit. This is how THREE.GLTFLoader works as well.
4. Don't automatically fetch the decoder source files (which might slow down page loading), but allow users to do that themselves if they prefer.

Updated usage example:

```js
// (Optional) Change decoder source directory (defaults to './').
THREE.DRACOLoader.setDecoderPath('./path/to/decoder/');

// (Optional) Use JS decoder (defaults to WebAssembly if supported).
THREE.DRACOLoader.setDecoderConfig({type: 'js'});

// (Optional) Pre-fetch decoder source files (defaults to load on demand).
THREE.DRACOLoader.getDecoderModule();

var dracoLoader = new THREE.DRACOLoader();
```

### DIFF
--- a/javascript/example/DRACOLoader.js
+++ b/javascript/example/DRACOLoader.js
@@ -353,18 +353,26 @@ THREE.DRACOLoader.setDecoderPath = function ( path ) {
 THREE.DRACOLoader.setDecoderConfig = function ( config ) {
   var wasmBinary = THREE.DRACOLoader.decoderConfig.wasmBinary;
   THREE.DRACOLoader.decoderConfig = config || {};
-  THREE.DRACOLoader.decoderModulePromise = null;
+  THREE.DRACOLoader.releaseDecoderModule();
 
   // Reuse WASM binary.
   if ( wasmBinary ) THREE.DRACOLoader.decoderConfig.wasmBinary = wasmBinary;
 };
 
- /**
-  * Gets WebAssembly or asm.js singleton instance of DracoDecoderModule
-  * after testing for browser support. Returns Promise that resolves when
-  * module is available.
-  * @return {Promise<{decoder: DracoDecoderModule}>}
-  */
+/**
+ * Releases the singleton DracoDecoderModule instance. Module will be recreated
+ * with the next decoding call.
+ */
+THREE.DRACOLoader.releaseDecoderModule = function () {
+  THREE.DRACOLoader.decoderModulePromise = null;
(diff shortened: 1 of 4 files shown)


donmccurdy:
See https://github.com/mrdoob/three.js/pull/12895; not sure which is the official copy.

javascript/example/DRACOLoader.js
```diff
@@ -346,131 +333,109 @@ THREE.DRACOLoader.prototype = {
...
+  // Reuse WASM binary.
+  if ( wasmBinary ) THREE.DRACOLoader.decoderConfig.wasmBinary = wasmBinary;
+};
 
-})();
-
-// Releases the DracoDecoderModule instance associated with the draco loader.
-// The module will be automatically re-created the next time a new geometry is
-// loaded by THREE.DRACOLoader.
-THREE.DRACOLoader.releaseDecoderModule = function() {
```
ondys: is there a reason why this got removed?

donmccurdy: Releasing now happens automatically if you call `DRACOLoader.setDecoderConfig({...})` and I assumed it wasn't needed otherwise, but could add an explicit method back if that's preferred.

donmccurdy:
There's also one more (small) change to DRACOLoader in https://github.com/donmccurdy/three.js/pull/6/, to resolve a race condition when an asset contains multiple Draco meshes.

javascript/example/DRACOLoader.js
donmccurdy: From https://github.com/google/draco/issues/312 I guess we do want to keep the method to explicitly release resources?

ondys: I think we should keep it there. Otherwise other changes look good to us.

donmccurdy: Ok, sounds good to me. The `.releaseDecoderModule()` method is back again.

ondys:
thanks for the PR!

Issue #319, opened 2017-12-22
gcc may be failing


FrankGalligan:
See https://travis-ci.org/google/draco/jobs/320412925

PR #320, opened 2017-12-23
Improve CMake C/CXX/AS flag handling.

tomfinegan:
Convert the flag controlling functions to macros and omit the
FORCE argument from all the set() calls. The use of FORCE was
mainly intended to set the variable in all scopes, but using
macros instead of functions is a much cleaner method of achieving
the same goal.


tomfinegan:
Poking around at #319... Let's see what Travis thinks.

PR #324, opened 2017-12-27
Enable static msvc runtime builds if MSVC_RUNTIME=static

pps83:
Currently draco is always build with MTd (multrithreaded dll runtime) on Visual Studio, which makes it difficult to use it in projects that require MT runtime. This change enables runtime override when running cmake with an extra param: `-DMSVC_RUNTIME=static`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,12 +168,15 @@ else ()
     endif ()
   endif ()
 
-  if (ENABLE_TESTS)
-    if (MSVC)
-      # Default runtime selected by cmake collides with Googletest settings,
-      # just force all Draco builds to be static (instead of dll/msvcrt).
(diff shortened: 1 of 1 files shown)


tomfinegan:
Apologies, I forgot to update this when it was addressed in another manner-- the msvc runtime issues run into w/combining Draco and GoogleTest were dealt with by dropping use of the GoogleTest CMake support in favor of using our own internal target.

tomfinegan:
Spoke too soon: not ready to publish this fix. The best option if in this situation is to disable tests (for now). Reopening this so it's not forgotten.

tomfinegan:
This has been obsoleted by the rebuilt CMake support that was released with v1.4. If the problem remains please create a new pull request.

PR #332, opened 2018-01-16
DRACOLoader: Support requests for multiple files in parallel

donmccurdy:
Calling `decodeDracoFile` currently caches the `attributeUniqueIdMap` on the instance, meaning:

- call `decodeDracoFile` with attribute map
- async request (1) begins
- call `decodeDracoFile` with a new attribute map, overriding `this.attributeUniqueIdMap`
- async request (2) begins
- async request (1) completes, begins decoding using the wrong attribute map

In this situation there will be issues in glTF assets containing multiple meshes. See https://github.com/donmccurdy/three.js/pull/6 for full context.


FrankGalligan:
Looks good. Thanks

PR #382, opened 2018-05-02
Accelerating Unity split mesh.

dyangrun:
I'm trying to decode 10 million mesh face, which is very slow.But not of Draco, it was slow to split mesh in Unity, so I did the unity split mesh optimization.And with multiple threads, there is almost no overhead for the main thread.
200s->8s


dyangrun:
I signed it!

FrankGalligan:
Thank you so much! The speed up is really impressive!

PR #385, opened 2018-05-05
THREE.DRACOLoader: Support attribute type map.

donmccurdy:
See https://github.com/mrdoob/three.js/pull/14002.


FrankGalligan:
Looks good. Thanks @donmccurdy

Issue #393, opened 2018-05-18
macos/ios unity build broken


tomfinegan:
Apparently clang doesn't like SOVERSION properties for bundle targets.

PR #394, opened 2018-05-18
Don't set SOVERSION property for ios/macos unity bundle.

tomfinegan:
It causes a link error like the following:
clang: error: invalid argument '-compatibility_version 1.0.0' only allowed with '-dynamiclib'

Fixes: https://github.com/google/draco/issues/393


tomfinegan:
This came up in a comment on https://github.com/google/draco/commit/3b3cab31e5e75de851cd2eb588cd7ba8eec70f27

PR #405, opened 2018-05-25
Add a flag for cmake to generate IE11 compatible JS decoder modules

ggetz:
The current JavaScript decoder module fails when run in IE11. 

This allows an `IE_COMPATIBLE` variable to be set when generating the cmake files in order to build a module that works in IE11 by providing polyfills, at a small hit to module size (1MB), with the solution proposed in https://github.com/kripken/emscripten/issues/6204


ggetz:
I've signed the CLA

PR #408, opened 2018-06-01
Get correct name from MetadataQuerier::GetEntryName

simenandresen:
This should fix a bug in GetEntryName which occures when e.g. trying to
get the name of each sub object in a mesh.

Previously, one would get the wrong name from a given `entry_id` since
one just looked up the name by index from the `metadata.entries_`. Since
`metadata.entries_` is just an unordered map, one would not get the
correct name.


simena86:
Would be nice if anyone could have a look here. Notice that there's not been made any changes to `draco_decoder_gltf` related files, since they were not built when following the build instructions
found [here](https://github.com/google/draco#webassembly-decoder)

simena86:
Not sure what's the best fix for the problem here, but IMO it's quite obvious that this is a bug. 

Have a look at the above diff, the original version loops over `metadata.entries()` which is an unordered map. Each entry is passed to a vector `entry_names_` which is later used to look up by index. Since each entry from `metadata.entries()` is passed into a vector randomly, there's no way this can work as far as I can see.

YouYue123:
I have made the same change. This indeed solve the problem to retrieve meta info in decoding process. Hope it can be merged

src/draco/javascript/emscripten/decoder_webidl_wrapper.cc
```diff
@@ -91,11 +91,15 @@ long MetadataQuerier::NumEntries(const Metadata &metadata) const {
...
                                           int entry_id) {
   if (entry_names_metadata_ != &metadata) {
-    entry_names_.clear();
+    entry_names_.resize(metadata.num_entries());
     entry_names_metadata_ = &metadata;
     // Initialize the list of entry names.
     for (auto &&entry : metadata.entries()) {
-      entry_names_.push_back(entry.first);
+      int name_index = -1;
+      entry.second.GetValue(&name_index);
```
ondys: wouldn't this assume that the metadata entry is of type "int" .. what if the data has a different type?

ram-sellamuthu:
> Not sure what's the best fix for the problem here, but IMO it's quite obvious that this is a bug.
> 
> Have a look at the above diff, the original version loops over `metadata.entries()` which is an unordered map. Each entry is passed to a vector `entry_names_` which is later used to look up by index. Since each entry from `metadata.entries()` is passed into a vector randomly, there's no way this can work as far as I can see.

I was testing your fix by downloading the javascript built and committed by you . however, I am still facing the issue where a sub-object with name "name" is shown. Kindly help.

FrankGalligan:
The code changed to use a map, so this should not be an issue anymore.

PR #424, opened 2018-06-15
Fix plugin build conditions

tomfinegan:
Build Maya and Unity plug-in libraries only when requested.


tomfinegan:
Obviated by other work.

PR #438, opened 2018-07-25
Unity Support Submesh And Support Encode Draco File

dyangrun:
Because of i use this, so simple support, only tested in Windows 7. 
 if you need a more sophisticated solution, you can modify this PR to add support


FrankGalligan:
Old PR, that no one has commented on.

Issue #458, opened 2018-09-01
Error: "Invalid number of indices on a face."


JohnnyPosi:
I'm getting this error when I'm trying to decode my `.obj` model. Is there any way to fix it?

JohnnyPosi:
Solved it: when export the model in Blender, just enable "Triangulate Faces" option. It increased the object size though.

jeromeetienne:
Just for logging, the same happened to me and the same solution worked. What about appending this message with 'Maybe try to triangulate your model?"

awong1900:
why do not support the Quad ?

FrankGalligan:
Closing as we added support to triangulate quads a while ago. See https://github.com/google/draco/blob/master/src/draco/io/obj_decoder.cc#L404

PR #460, opened 2018-09-05
Added suggestion on error message for .obj

jeromeetienne:
Obj can create a single face with a lot vertices (e.g. in Maya export). obj_decoder.cc support only 4 vertices per face. if a face got more than 4 vertices, a message "Invalid number of indices on a face" is displayed in console.

I added a suggestion to let the user know how to fix it.

Related to #458 

PS: great work on draco, the output size is impressive, thanks for doing the hard work and making it opensource


jeromeetienne:
hmm i didnt remember the CLA thing. not that my contributions is significative anyway :)

thanks for making draco available, keep on the good work i love it

PR #473, opened 2018-10-06
Allow to compile with Draco without warnings

thibault-p:
Remove extra semicolons after macros which are reported as warnings when compiling with `pedentic` flag.


thibault-p:
I signed it!

PR #486, opened 2018-12-06
animation/keyframe_animation: Use unsigned values instead for variables that store quantities

FreddyFunk:
Quantities should never be negative. Therefore we should use unsigned values instead of signed values to store them.


FrankGalligan:
In general I agree that these values could have been unsigned. Actually they should probably have been size_t. But changing the api at this point could be a huge issue.

PR #487, opened 2018-12-06
attributes/normal_compression_utils: Fix compiler warnings and refactor template functions

FreddyFunk:
This PR fixes 27 compiler warnings (MSVC) due to possible lossy type conversions and simplifies some nested and overcomplicated if-statements.


FrankGalligan:
We changed the code internally and there should not be warning anymore.

PR #488, opened 2018-12-06
point_cloud: Fix compiler warning

FreddyFunk:
This PR adds a missing cast to eliminate a compiler warning.


FrankGalligan:
Thanks!

PR #521, opened 2019-05-16
Add MetadataQuerier.GetIntEntryArray to JavaScript bindings

eile:
We need this to read some internal metadata on our geometries. If possible, we would like to get this deployed through an official release soonish.


ondys:
Thanks for the PR

PR #531, opened 2019-06-12
TravisCI: Use newer linux images.

tomfinegan:
Update the linux jobs to use Xenial. This simplifies our setup: We no longer need to check for the presence of packages other than ccache and cmake.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,27 +3,17 @@ language: cpp
 matrix:
   include:
     - os: linux
-      dist: trusty
+      dist: xenial
       compiler: clang
     - os: linux
-      dist: trusty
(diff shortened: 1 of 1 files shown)


tomfinegan:
This just needs a rubber stamp. It clears the strange failures in the clang build. Example here:
https://api.travis-ci.org/v3/job/544202129/log.txt

Somehow -Werror is getting passed in CXX flags, and unused include paths are setting it off:

> clang-5.0: warning: argument unused during compilation: '-I /home/travis/build/google/draco/../googletest'

I can't repro this locally (on a debian machine using clang-5.0), and moving to xenial images seems to solve the problem.

PR #538, opened 2019-07-03
Fix: Unity requires static libs for iOS.

atteneder:
Changed the library type from MODULE to STATIC for dracodec_unity IOS only. For other platforms it stays MODULE.

Tested with CMake 3.14.5 (has built-in iOS toolchain)

```bash
cmake .. -G Xcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES=armv7\;armv7s\;arm64 -DCMAKE_OSX_DEPLOYMENT_TARGET=10.0 -DCMAKE_XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH=NO -DBUILD_UNITY_PLUGIN=ON
```


atteneder:
I signed it!

PR #540, opened 2019-07-10
Improved Unity integration and package

atteneder:
Improvements

- Is magnitudes faster due to
  - Bulk memory copies instead of per vertex/index data copy
  - Multi-threaded via C# Job system
- Supports single meshes with more than 65536 vertices ( Unity 2017.3 and newer only; old split algorithm was broken anyways)
- Added/fixed platforms (without native libs in this PR)
  - iOS
  - WebGL

I took the existing Unity scripts and improved them in a separate Repository, since I wanted users to be able to easily include the Draco library via Unity package manager. The new repository with the complete commit history can be found here: <https://github.com/atteneder/DracoUnity>. There's also a [Demo project](https://github.com/atteneder/DracoUnityDemo) that uses it.

Additional changes the separate repository brings:
- Can be integrated into Projects easily via Package Manager
- Builds of native library for all missing platforms (no need for users to build draco themselves)
  - WebGL
  - iOS armv7(s) and arm64
  - Windows 32-bit
  - Linux 64-bit and 32-bit
  - Android x86
- Existing libs got updated as well
  - Android arm
  - Windows 64-bit
  - macOS 64-bit

So this PR is a copy / back-port of the work above and somehow a placeholder. I'd be glad if you agree that a separate unity package/repository makes sense and adopt it as well.


FrankGalligan:
Hi @atteneder Thanks for all of your work. I would like to accept this, but our knowledge of Unity is extremely limited. 

In the readme we [copy over the scripts](https://github.com/google/draco/tree/master/unity#copy-unity-script-to-your-project). Then we [attach ](https://github.com/google/draco/tree/master/unity#load-draco-assets-in-runtime)the DracoDecodingObject.cs to the game object.

With this PR you remove DracoDecodingObject.cs. What are the steps to get an encoded Bunny to render in the simple project?

Thanks

atteneder:
Hi Frank,

good to hear you're considering this!

This PR shouldn't be merged directly. It's a placeholder for the idea of making the Unity integration a Unity Package (for Unity's Package Manager). If you're familiar with node/npm, the concept is largely the same. It's easier for developers to integrate and update the Draco "package" in their Unity projects.

You can find the most recent version of my port and install instructions here:
<https://gitlab.com/atteneder/DracoUnity#installing>

To learn more about Unity packages:
<https://docs.unity3d.com/Manual/CustomPackages.html>

There are several ways of sharing/distributing Unity packages. As I am an individual coder and not a company, I chose via a GIT repository, but there are other options:
<https://docs.unity3d.com/Manual/cus-share.html>

Hope that helps. Let me know if you have trouble with it.

Cheers

FrankGalligan:
OK got it. So what I was thinking was to keep the super simple example we have now for Unity on Draco github. Then basically have a bunch of links to your website for more information on working with Draco and Unity. Do you mind if we link to https://gitlab.com/atteneder/DracoUnity as being the better authority of using Draco with Unity?


Then I will probably pull in the performance enhancements and bug fixes you made in this PR and leave the package code out.

atteneder:
Sounds like a solution that works for all cases (quick and simple test run and thourough integration via link).
Feel free to add the link!

Thanks

FrankGalligan:
Thanks

atteneder:
I suggest to use the master branch of my fork.

this unity_on_steroids branch is outdated and doesn not contain:
- GitHub actions builds for all platforms
- Draco to Unity space conversions
- Tangent fix (although I'm onto making this optional for performance reasons)

FrankGalligan:
OK, I made changes upstream just to point to here:
https://gitlab.com/atteneder/DracoUnity

Which will be on the master branch by default.

atteneder:
Thanks!
I guess we can close this PR then? Feel free to do so!

FrankGalligan:
We pushed a new release which has the faster conversion functions (a little different than this PR) in this pull request.

You can see it defined here:
https://github.com/google/draco/blob/master/src/draco/javascript/emscripten/draco_web_decoder.idl#L259

You can see an example of use here:
https://github.com/google/draco/blob/master/javascript/example/DRACOLoader.js#L575

Also now for Unity we default to point here:
https://gitlab.com/atteneder/DracoUnity

See text in README:
https://github.com/google/draco#unity

shendibrani:
Trying to render the bunny on Unity using the package is failing to decode the draco file.

Issue #471, opened 2018-10-02
EncodeGoldenMesh test failure with e9337d1faaa667d313e376d48dbc7007254fb032


hobu:
Using MSVC 2015, I am getting a failure with EncodeGoldenMesh

```
[ RUN      ] MeshEncoderTests/MeshEncoderTest.EncodeGoldenMesh/0
Test output size differed from golden file sizeC:\projects\draco\src\draco\compression\mesh\mesh_encoder_test.cc(78): error: Value of: CompareGoldenFile(golden_file_name, buffer.data(), buffer.size())
  Actual: false
Expected: true
Encoded data is different from the golden file. Please verify that the encoding works as expected and update the golden file if necessary (run the test with --update_golden_files flag).
[  FAILED  ] MeshEncoderTests/MeshEncoderTest.EncodeGoldenMesh/0, where GetParam() = "sequential" (12 ms)
[ RUN      ] MeshEncoderTests/MeshEncoderTest.EncodeGoldenMesh/1
Test output size differed from golden file sizeC:\projects\draco\src\draco\compression\mesh\mesh_encoder_test.cc(78): error: Value of: CompareGoldenFile(golden_file_name, buffer.data(), buffer.size())
  Actual: false
Expected: true
Encoded data is different from the golden file. Please verify that the encoding works as expected and update the golden file if necessary (run the test with --update_golden_files flag).
[  FAILED  ] MeshEncoderTests/MeshEncoderTest.EncodeGoldenMesh/1, where GetParam() = "edgebreaker" (10 ms)
[----------] 2 tests from MeshEncoderTests/MeshEncoderTest (25 ms total)

[----------] Global test environment tear-down
[==========] 143 tests from 33 test cases ran. (1009 ms total)
[  PASSED  ] 141 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] MeshEncoderTests/MeshEncoderTest.EncodeGoldenMesh/0, where GetParam() = "sequential"
[  FAILED  ] MeshEncoderTests/MeshEncoderTest.EncodeGoldenMesh/1, where GetParam() = "edgebreaker"
```

FrankGalligan:
I just checked with the forthcoming release and this works in MSVC 17.

PR #557, opened 2019-09-25
Fix unit test golden file reading (#471)

avzhatkin:
Apparently MSVC runtime in text mode interprets 0x1a as end of file. So binary mode is necessary for reading file.


avzhatkin:
@googlebot I signed it!

avzhatkin:
@googlebot I fixed it.

PR #559, opened 2019-09-27
Add vcpkg installation instructions

grdowns:
Draco is available as a port in [vcpkg](https://github.com/Microsoft/vcpkg), a C++ library manager that simplifies installation for Draco and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build Draco, ready to be included in their projects. 

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/draco/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,18 @@ _**Contents**_
 
 Building
 ========
-For all platforms, you must first generate the project/make files and then
+
+You can download and install Draco using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    vcpkg install draco
+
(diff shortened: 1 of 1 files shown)


tomfinegan:
Thanks for the contribution, and for the the addition to vcpkg.

I'm not going to merge this as-is. I would be willing to consider adding this to the readme in a later section (perhaps an addendum/appendix that covers packagers and package management support), but listing it as if it's the intended and supported build environment by the Draco project is not desirable as the project has no connection to vcpkg and has no dependency on it.

FrankGalligan:
Added vcpkg info here:
https://github.com/google/draco/blob/master/BUILDING.md#vcpkg

Issue #585, opened 2020-03-08
Link in README not working


md2perpe:
The link https://www.gstatic.com/draco/v1/decoders/ found early in README.md results in a 404 error.

FrankGalligan:
I put up a change to try and better explain that there are a few files at https://www.gstatic.com/draco/v1/decoders/, with a concrete example of pulling https://www.gstatic.com/draco/v1/decoders/draco_decoder_gltf.wasm

PR #587, opened 2020-03-09
Update README with more info on gstatic link

FrankGalligan:


FrankGalligan:
This addresses issue #585

PR #601, opened 2020-04-29
Additions to JS decoder API

jvail:
I'd like to propose some additions to the JS decoder API (hope I haven't missed something in the API that makes these changes redundant):

* add GetDoubleEntryArray to MetadataQuerier
* add GetBinaryEntryArray to MetadataQuerier
* add data() method to DracoArray

I am storing arrays as metadata and I was looking for an easy way to copy the data over form the emscripten heap rather than iterating:

```
const arr = new draco.DracoFloat32Array();
metaQuerier.GetDoubleEntryArray(metadata, name, arr);
const size = arr.size();
const ptr = arr.data().ptr;
```

Then e.g.
```
entry = new Float32Array(draco.HEAPU8.buffer.slice(ptr, ptr + size * 4));
```

instead of 
```
entry = new Float32Array(size);
for (let idx = 0; idx < size; idx++) {
  entry[idx] = arr.GetValue(idx);
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 docs/_site
+build
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 docs/_site
-build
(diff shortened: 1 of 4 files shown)


jvail:
@googlebot I signed it!

jvail:
Hello? Is there anyone else except the bot? Is the PR nonsense or will you consider it?

PR #626, opened 2020-09-10
Add version string output to cli tools.

naxostech:
Add -v arg that causes the tools to print version and exit when specified.


tomfinegan:
@naxostech Can you push this again, and sign the CLA? I would like to merge this. 

Thanks!

PR #630, opened 2020-09-21
Uses SVG logo in README

gsimone:
Super minor, but I was looking at the repo and noticed the logo was blurry, poked around a bit and found the svg version already in the repo, so why not try and use it for the README?


gsimone:
@googlebot I signed it!

Issue #467, opened 2018-09-25
C-API for FFI


SirVer:
I'd like to use draco in https://github.com/googlecartographer/point_cloud_viewer, but I am not using C++. Binding to the library would be much easier if there was a minimal C-API. This would also allow to use the library easily from Python.

Is it feasible to add a (minimal) C-API, tailored for FFI?

AlexApps99:
Hello, I'm currently making a draco binding for Rust and a C binding would help me trememdously.
I'm sure that it would help with many other similar projects too, such as a Python binding and C game engines.

qmuntal:
I would also be very helpful for Go interop. I´m specially interested to support draco compression in https://github.com/qmuntal/gltf.

This issue is marked as `help wanted`, so I will give it a try during the following weeks.

yw13931835525-cyber:
河北高软科技有限公司愿意承接此项目。

公司优势：
- 61个专业Agent团队
- 7x24小时开发
- AI驱动，成本优势
- 有丰富的C/C++ FFI绑定和Python ctypes开发经验

技术方案：
1. 设计最小化的C-API接口（encoder/decoder init, encode, decode, release）
2. 使用纯C实现，不依赖C++标准库，便于FFI
3. 提供Python ctypes绑定示例
4. 编写完整的单元测试和文档

报价：$2,000 - $3,500
工期：3-4周

收款：USDT(ERC20) 0x6FCBd5d14FB296933A4f5a515933B153bA24370E
或对公账户（河北高软科技有限公司）

PR #663, opened 2020-12-14
Initial C API for FFI

qmuntal:
## Goal

This PR contains a proposal of C API to facilitate the interoperability with languages other than C++.
It does not try to define and implement an API for all the C++ API, just a minimal decoding subset, as there are a lot of design decisions that need to be made before going for the full implementation. 

##  Context

- There is interest in having a plain C interface to interop with other languages. See #467.
- There is no official C style guide documented in the repo.
- There are already three draco wrappers implemented in this repo:
  - Unity: Defines a quite complete C-like header using some C++ features, such as struct default values and namespaces. It 
 imports other C++ headers from the `src` tree. Distributed in a separate library.
  - Maya: Defines a simple C++ header importing other C++ headers from the `src` tree. Distributed in a separate library.
  - Ecmascript: Defines a a complete WebIDL interface which can be distributed as a JS or a WASM module.

## Design Goals

- Allow creation of bindings for other languages through typical C foreign function interfaces
- Avoid language features that raise barrier to use. E.g. no passing of structs by-value or post-C99 features
- Achieve link-time compatibility between different implementations. E,g, no struct definitions in the API

## Decision points

- The new API will be distributed as part of the core library
  - Facilitate distribution as there is only one library to share 
  - C API will be always aligned with the wrapped code, no possibility of version mismatch.
- No reuse of any already implemented wrapper (unity, maya and ecmascript)
  - These libraries do not define a plain C API
  - Unity and maya plugins serve completely different objectives
- C API gated behind `DRACO_C_API` cmake flag. Disabled by default.
- Object oriented C API using opaque pointers with a unique incomplete type for each object (no `void*`)
  - Hides implementation details
  - User friendly
  - Draco objects memory management is owned by the C API implementation
  - `void *` removes all semblance of type safety
  - Achieve link-time compatibility even when wrapped structs memory layout changes
 
## Example of FFI

I´ve implemented a Go wrapper using the C API provided in this PR to demonstrate its capabilities. You can find it in https://github.com/qmuntal/go-draco/blob/master/draco.go.

## Next steps

- Iterate on this PR until the draco team is happy with the C API look and feel.
- Incrementally add new methods to the C API in further PRs.

### DIFF
--- a/src/draco/c_api/c_api.h
+++ b/src/draco/c_api/c_api.h
@@ -1,0 +1,95 @@
+// Copyright 2020 The Draco Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef DRACO_C_API_H_
+#define DRACO_C_API_H_
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// If compiling with Visual Studio.
+#if defined(_MSC_VER)
+#define EXPORT_API __declspec(dllexport)
+#else
+// Other platforms don't need this.
+#define EXPORT_API
+#endif  // defined(_MSC_VER)
+
+typedef const char* draco_string; // NULL terminated
+
+// draco::Status
+
+typedef struct draco_status draco_status;
+
+EXPORT_API void dracoStatusRelease(draco_status *status);
+
+EXPORT_API int dracoStatusCode(const draco_status *status);
+
+EXPORT_API bool dracoStatusOk(const draco_status *status);
+
+// Returns the status message.
+// The memory backing memory is valid meanwhile status is not released.
+EXPORT_API draco_string dracoStatusErrorMsg(const draco_status *status);
+
+// draco::Mesh
+
+typedef uint32_t draco_face[3];
+typedef struct draco_mesh draco_mesh;
+
+EXPORT_API draco_mesh* dracoNewMesh();
+
+EXPORT_API void dracoMeshRelease(draco_mesh *mesh);
+
+EXPORT_API uint32_t dracoMeshNumFaces(const draco_mesh *mesh);
+
+EXPORT_API uint32_t dracoMeshNumPoints(const draco_mesh *mesh);
+
+// Queries an array of 3*face_count elements containing the triangle indices.
+// out_values must be allocated to contain at least 3*face_count uint16_t elements.
+// out_size must be exactly 3*face_count*sizeof(uint16_t), else out_values
+// won´t be filled and returns false.
+EXPORT_API bool dracoMeshGetTrianglesUint16(const draco_mesh *mesh,
+                                            const size_t out_size,
+                                            uint16_t *out_values);
+
+// Queries an array of 3*face_count elements containing the triangle indices.
+// out_values must be allocated to contain at least 3*face_count uint32_t elements.
+// out_size must be exactly 3*face_count*sizeof(uint32_t), else out_values
+// won´t be filled and returns false.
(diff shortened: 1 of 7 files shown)


src/draco/c_api/c_api.h
```diff
@@ -0,0 +1,95 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// If compiling with Visual Studio.
+#if defined(_MSC_VER)
+#define EXPORT_API __declspec(dllexport)
+#else
+// Other platforms don't need this.
```
RichardTea, line 29: GCC and Clang do need an EXPORT_API definition since GCC v4
It is strongly recommended not to export all functions, especially for a C-API as it greatly reduces the chance of name collisions and speeds up the dylib/so loader.

The draco C-API so/dylib should be compiled with the flag -fvisibility=hidden, and the following added for GCC and clang compilers:
```
#ifdef DRACO_BUILDING_DLL // also update draco_targets.cmake
#define EXPORT_API __attribute__ ((visibility ("default")))
#else
#define EXPORT_API
#endif
```

src/draco/c_api/c_api.h
```diff
@@ -0,0 +1,95 @@
...
+#include <stdint.h>
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// If compiling with Visual Studio.
+#if defined(_MSC_VER)
+#define EXPORT_API __declspec(dllexport)
```
RichardTea, line 27: Should be:
```
#ifdef DRACO_BUILDING_DLL
#define EXPORT_API __declspec(dllexport)
#else
#define EXPORT_API __declspec(dllimport)
#endif
```

RichardTea (changes requested):
The c_api.h header should be set up for use by consuming applications as well as the draco library itself.

Additionally, draco's C-API should _not_ export all symbols, it should only export the symbols required for the C API

src/draco/c_api/c_api.h
qmuntal: Done!

src/draco/c_api/c_api.h
qmuntal: This is a little bit tricky to implement with the current approach of building the C-API as part of the "core" library, but I do agree it should happen.

The problem I'm facing if only the C-API is exported then the executables (`draco_decoder` and `draco_encoder`) nor the test can be correctly linked as those depend on the C++ API.

I'm starting to lean towards having a separate library for the C-API, as the CMake orchestration would be much more easy to manage and the `DRACO_C_API` will be orthogonal to other features. What do you think @RichardTea?

RichardTea: I think you're right. While I'm sure it _possible_ to build all four, it'd be quite complex.

I think it's reasonable to pick an API for your shared libraries at configure time, as I'm reasonably sure that most consumers of the C-API will _not_ want the C++ version of the dynamic library, so no need to jump through the hoops to make them both happen.

I _think_ draco_encoder, draco_decoder and the C-API shared lib could all consume the same C++ static library "privately" to achieve the goal of hiding all the internal symbols - IIRC, symbols aren't ever truly private in a static library so that should still work.

qmuntal: I've finally moved the C-API to its own library called `cdraco.[dll/lib/so/a/...]`, gated by the `DRACO_C_API` flag. This library is self contained and the dynamic version just exports the C-API, all the C++ stuff is hidden.

qmuntal:
I'm adding a new optional pre-processor definition: `DRACO_STATIC`. It must be set by `cdraco` consumers that are statically linking in windows. It is unfortunate but required, as there is no automatic (that I'm aware of) to not use the `__declspec(dllimport)` when statically linking, and this causes the linker to search for function names with the harcoded `__imp_´ prefix.

tomfinegan:
@qmuntal: Thank you for the preliminary design and patches for adding a C-API to Draco.

At this time the Draco team does not have the resources to accept a C-API. We are not currently able to review, test, or maintain the additional API and glue code necessary for a fully functional C-API.

If a C-API is a critical requirement for your purposes, at least for the time being, you'll need to maintain it in a fork of the Draco repository, or in another project that depends on the Draco repository.  

Again, thank you for the PR, but we cannot accept it into the project at present. I will close this PR for now. If the situation changes the Draco team will revisit this subject.

qmuntal:
@RichardTea @tomfinegan I completely understand the situation and appreciate the efforts you have put on this reviews.

I'll probably keep this changes in a fork in order to support at least the Go wrapper I plan to implement (https://github.com/qmuntal/go-draco).

And of course I'll be more than happy to help on revamping this PR if your team have enough bandwidth to maintain it.

PR #677, opened 2021-02-03
Support building mingw static and dynamic at the same time

qmuntal:
Mingw-w64 (and Cygwin) uses the following extension schema for generated libraries:
- Shared library: `.dll`
- Import library:  `.a.dll`
- Static library: `.a`

Notice that there is no basename collision between shared and static libraries, so it is safe to support generating both types at the same time.

The basename collision issue happens only in MSVC (to the best of my knowledge).

### DIFF
--- a/cmake/draco_targets.cmake
+++ b/cmake/draco_targets.cmake
@@ -163,7 +163,7 @@ endmacro()
 # cmake-format: off
 #   - OUTPUT_NAME: Override output file basename. Target basename defaults to
 #     NAME. OUTPUT_NAME is ignored when BUILD_SHARED_LIBS is enabled and CMake
-#     is generating a build for which MSVC or WIN32 are true. This is to avoid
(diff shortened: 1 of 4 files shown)


cmake/draco_targets.cmake
```diff
@@ -163,7 +163,7 @@ endmacro()
 # cmake-format: off
 #   - OUTPUT_NAME: Override output file basename. Target basename defaults to
 #     NAME. OUTPUT_NAME is ignored when BUILD_SHARED_LIBS is enabled and CMake
-#     is generating a build for which MSVC or WIN32 are true. This is to avoid
+#     is generating a build for which MSVC are true. This is to avoid
```
tomfinegan, line 166: nit: are -> is

tomfinegan (changes requested):
Looks fine. One minor nit on one of the comment changes.

PR #690, opened 2021-03-04
Unity related changes

atteneder:
Hello dear Draco developers,

This PR mostly serves the purpose of suggesting additions/changes, kick off discussions and not primarily to be merged in the end.

The code can be tested in the related Unity package [DracoUnity 2.0.0-preview](https://github.com/atteneder/DracoUnity).

## Split up decode function

The decode function could be sped up for Unity (and maybe other systems as well) by splitting it into phases:

1. Parse Draco data and provide final index and vertex buffer structure and size. Unity is then able to allocate resources properly ahead of time (which has to be done on the main thread)
2. Perform actual decoding. This can now be done efficiently on threads, even in parallel (for each vertex attributes)

## Encoding

I've created an additional library that allows encoding Unity meshes to draco data/files. 

## Others

- GitHub action to build Unity libraries for all platforms in one go
- Optional coordinate system conversion (required in glTF context)
- Provide Unity package for easy installation/usage (see [DracoUnity](https://github.com/atteneder/DracoUnity))

The latter is nothing new, but I'm open to re-integrate this package into the original repository. We could create a GitHub action that builds the native libaries, adds it to the package and deploys it to a special branch. From there it could be deployed to a package manage server (like verdaccio or OpenUPM)

Let me know your thoughts,
Thanks


tomfinegan:
@atteneder Thanks for the PR. I'm going to start by going over the GitHub actions and CMake stuff. Frank and l will look more deeply into the code over the next week or so-- things are pretty busy at present.

Re the GH actions stuff: there are some internal policy items that necessitate some changes to our repo configuration before we can accept CI automation that touches third-party actions. This shouldn't impact you much, but it's slowing the process down because Google projects can't accept the addition of third-party actions without these changes. As I currently understand the only change on your side is that CI integration must use specific released versions of third-party actions, but beyond that nothing is jumping out at me.

.github/workflows/unity.yml
```diff
@@ -0,0 +1,249 @@
...
+
+
+jobs:
+
+  mac:
+
+    runs-on: macOS-latest
+
+    steps:
+    - uses: actions/checkout@v1
```
tomfinegan, line 23: Why v1 here?

.github/workflows/unity.yml
```diff
@@ -0,0 +1,249 @@
...
+
+    - uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: latest-stable
+
+    - uses: seanmiddleditch/gha-setup-ninja@master
+
+    - name: configure_mac
+      # ARCHS_STANDARD explicitely enables Apple silicon on Xcode 12.2
+      run: cmake . -G Ninja -B build_mac -DCMAKE_BUILD_TYPE=MinSizeRel -DCMAKE_OSX_ARCHITECTURES="x86_64;arm64" -DDRACO_UNITY_PLUGIN=ON -DDRACO_GLTF=ON -DDRACO_BACKWARDS_COMPATIBILITY=OFF
```
tomfinegan, line 33: Here and below: Please wrap these long commands if Github's YAML setup allows.

.github/workflows/unity.yml
```diff
@@ -0,0 +1,249 @@
...
+    runs-on: macOS-latest
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: latest-stable
+
+    - uses: seanmiddleditch/gha-setup-ninja@master
```
tomfinegan, line 29: Can you pin this to a tag? It's probably harmless, but my preference is to at least reference a release here. I think v3 will work.

.github/workflows/unity.yml
```diff
@@ -0,0 +1,249 @@
...
+        name: draco_win
+        path: draco_win
+
+
+  linux:
+
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v1
```
tomfinegan, line 126: Ditto v1 comment from above

.github/workflows/unity.yml
```diff
@@ -0,0 +1,249 @@
...
+    - name: upload artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: draco_win
+        path: draco_win
+
+
+  linux:
+
+    runs-on: ubuntu-18.04
```
tomfinegan, line 123: Why not use ubuntu-latest?

.github/workflows/unity.yml
```diff
@@ -0,0 +1,249 @@
...
+
+    steps:
+    - uses: actions/checkout@v1
+
+    # Probably not needed anymore with ubuntu 20.04
+    - name: install_cmake
```
tomfinegan, line 129: The existing draco action uses ubuntu-latest and omits this step-- I think you can drop the CMake installation.

.github/workflows/unity.yml
```diff
@@ -0,0 +1,249 @@
...
+      run: |
+        sudo apt update
+        sudo apt install apt-transport-https ca-certificates gnupg software-properties-common wget
+        wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | sudo apt-key add -
+        sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'
+        sudo apt update
+        sudo apt install cmake
+
+    - name: apt install
```
tomfinegan, line 138: I think you can drop this: gha-setup-ninja does this for you, no?

.github/workflows/unity.yml
```diff
@@ -0,0 +1,249 @@
...
+
+    - name: package Apple
+      run: |
+        mkdir draco_apple
+        mkdir draco_apple/x86_64
+        mv build_mac/dracodec_unity.bundle draco_apple/x86_64
+        mv build_mac/dracoenc_unity.bundle draco_apple/x86_64
+        mkdir draco_apple/iOS
+        mv build_ios/MinSizeRel-iphoneos/libdracodec_unity.a draco_apple/iOS
+    - name: upload artifact
```
tomfinegan, line 52: Generally speaking I'm fine with everything here accept the automated packaging and release. I think what we'll want to do here, if you're willing, is keep the CI stuff and omit the rest. The additional coverage is great, but the packaging and release automation stuff runs afoul of some of our open source policy.

CMakeLists.txt
```diff
@@ -490,6 +490,10 @@ list(APPEND draco_unity_plug_sources
             "${draco_src_root}/unity/draco_unity_plugin.cc"
             "${draco_src_root}/unity/draco_unity_plugin.h")
 
+list(APPEND draco_unity_enc_plug_sources
```
tomfinegan, line 493: Please run cmake-format on the CMake files. You can install 0.5.1 using pip via 'pip install cmake-format==0.5.1'.

CMakeLists.txt
```diff
@@ -882,13 +914,61 @@ else()
...
+                      draco_unity_enc_plugin
                       LIB_DEPS
                       ${draco_plugin_dependency})
 
     # For Mac, we need to build a .bundle for the unity plugin.
     if(APPLE)
       set_target_properties(dracodec_unity PROPERTIES BUNDLE true)
+      set_target_properties(dracoenc_unity PROPERTIES BUNDLE true)
+    endif()
+    if(IOS)
```
tomfinegan, line 940: It would be better update the flags for IOS in cmake/draco_build_definitions.cmake. Just add something like this to the bottom of draco_set_build_definitions().

```
if(IOS)
  list(APPEND draco_base_cxx_flags "-fembed-bitcode")
endif()
```

CMakeLists.txt
```diff
@@ -882,13 +914,61 @@ else()
...
+      target_compile_options( draco_dec_config PUBLIC -fembed-bitcode )
+      target_compile_options( draco_io PUBLIC -fembed-bitcode )
+      target_compile_options( draco_mesh PUBLIC -fembed-bitcode )
+      target_compile_options( draco_metadata PUBLIC -fembed-bitcode )
+      target_compile_options( draco_metadata_dec PUBLIC -fembed-bitcode )
+      target_compile_options( draco_point_cloud PUBLIC -fembed-bitcode )
+      target_compile_options( draco_points_dec PUBLIC -fembed-bitcode )
+      target_compile_options( draco_unity_plugin PUBLIC -fembed-bitcode )
+    endif()
+    if (WIN32 AND ${CMAKE_SYSTEM_NAME} STREQUAL "WindowsStore")
```
tomfinegan, line 961: Same with this one, but we already have MSVC specific flags. Just update the draco_msvc_cxx_flags list, and add /wd4146.

cmake/toolchains/i686-linux.cmake
```diff
@@ -0,0 +1,16 @@
...
+
+set(BIN_LIBROOT "linux")
+set(BIN_RPATH "\$ORIGIN/lib")
+
+set(CMAKE_SKIP_BUILD_RPATH TRUE)
+set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
+set(CMAKE_INSTALL_RPATH ${BIN_RPATH})
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH FALSE)
+
+set(FORCE32 ON)
```
tomfinegan, line 16: Is this needed?

cmake/toolchains/i686-linux.cmake
```diff
@@ -0,0 +1,16 @@
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_VERSION 1)
+set(CMAKE_SYSTEM_PROCESSOR "i686")
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -m32" CACHE STRING "c++ flags")
+set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -m32" CACHE STRING "c flags")
+
+set(BIN_LIBROOT "linux")
+set(BIN_RPATH "\$ORIGIN/lib")
```
tomfinegan, line 9: Does CMake populate $ORIGIN at some point?

tomfinegan (changes requested):
Sorry for the slow turnaround on this. I've really only gone through the CI and build related changes. Overall they look fine excepting the packaging related stuff, which I don't think we can accept in the parent repo.

.github/workflows/unity.yml
atteneder: I've had a users report an [issue](https://github.com/atteneder/DracoUnity/issues/8) when using a library built on ubuntu-latest (20.10 at the time) on ubuntu 18.04 (correct glibc version was not found).

Is there a better way to achieve backwards compatibility? Or cross distribution compatibility in general?

.github/workflows/unity.yml
atteneder: True.
This depends on switching to Ubuntu 20.04 or newer (conversation above), as it comes with a CMake>=3.12 (which is required)

.github/workflows/unity.yml
atteneder: Let me explain why I went the extra mile for this:

When building/updating the libraries for all platforms Unity developers have to copy each single file into the correct location within their project/package. If the structure is already like this in the CI result, this is simplified to a single download and unzip command instead of a lot of manual torque commands.

Fair enough though, If you could confirm that this still is not accepted I'll move it to a fork for personal use. And please feel free to suggest a better solution that would be accepted.

CMakeLists.txt
atteneder: Will do!
Note: cmake-format did changes in sections not touched by me. I'll only commit changes to lines I added/changed.

.github/workflows/unity.yml
atteneder: Clear mistake. changed to v2 everywhere

cmake/toolchains/i686-linux.cmake
atteneder: I removed i686 completely (it's not supported by Unity 2019.2 or later anyways)

atteneder:
Thank you very much @tomfinegan for the excellent feedback and sorry I wasn't able to pick it back up for so long.

I tried to implement all suggestions and returned some questions where things are not clear to me. Depending on your answer I'll remove the packaging (or not), test if the CI works and make the PR ready for review.

PR #693, opened 2021-03-16
Use latest available OS images in GitHub actions.

tomfinegan:


tomfinegan:
Overriding the review check because moving the GHA test runs to the latest OS images is trivial.

PR #698, opened 2021-04-13
[fuzzing] Fix build.sh

jonathanmetzman:
Fixes https://github.com/google/oss-fuzz/issues/5610

### DIFF
--- a/src/draco/tools/fuzz/build.sh
+++ b/src/draco/tools/fuzz/build.sh
@@ -19,7 +19,7 @@
 cmake $SRC/draco
 # The draco_decoder and draco_encoder binaries don't build nicely with OSS-Fuzz
 # options, so just build the Draco shared libraries.
-make -j$(nproc) draco
+make -j$(nproc)
 
 # build fuzzers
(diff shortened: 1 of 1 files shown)


src/draco/tools/fuzz/build.sh
```diff
@@ -19,7 +19,7 @@
 cmake $SRC/draco
 # The draco_decoder and draco_encoder binaries don't build nicely with OSS-Fuzz
 # options, so just build the Draco shared libraries.
```
tomfinegan, line 21: Please drop this comment about draco_decoder and draco_encoder if this is working now.

tomfinegan (changes requested):

src/draco/tools/fuzz/build.sh
jonathanmetzman: Actually I just tested this and it seems to fail for non-AFL builds still.
Since this is going to be a more involved fix, do you think you all can look at why the OSS-Fuzz build fails ([logs](https://oss-fuzz-build-logs.storage.googleapis.com/log-6f5f4441-af18-4dd2-877d-3d69c812964b.txt)).
Either way, I'll close this PR.

Issue #764, opened 2021-11-10
Incorrect install paths


jansol:
When given an absolute path for e.g. `CMAKE_INSTALL_LIBDIR`, cmake does NOT prepend `CMAKE_INSTALL_PREFIX`, that is only done for relative paths (see [the CMake manual](https://cmake.org/cmake/help/latest/command/install.html))

However path wrangling in `draco_install.cmake` always unconditionally prepends the prefix, leading to broken installs e.g. on [NixOS](https://github.com/NixOS/nixpkgs/pull/143782). Nix always passes absolute paths to cmake for the specific install directories, as they may have different prefixes in order to split large packages into smaller ones (such as having a separate package for documentation or data files).

tomfinegan:
> When given an absolute path for e.g. CMAKE_INSTALL_LIBDIR, cmake does NOT prepend CMAKE_INSTALL_PREFIX, that > is only done for relative paths (see the CMake manual)

Can you cite the section of the manual that you're referring to? I took a quick look over the script, and just reread the CMAKE_INSTALL_PREFIX documentation. There is no mention of behavioral changes relating to absolute paths. The GNUInstallDirs documentation is pretty clear that CMAKE_INSTALL_LIBDIR is not absolute and is built using CMAKE_INSTALL_PREFIX. CMAKE_INSTALL_FULL_LIBDIR is absolute, but the CMAKE_INSTALL_LIBDIR is not. Specifying an absolute path as the value of that variable appears incorrect by my reading. 

The CMake install() code in Draco is ancient, and I see from looking at the documentation that things have changed, but the problem you're having downstream is clearly related to the behavior of Nix, which is not typical:

> Nix always passes absolute paths to cmake for the specific install directories.

CMAKE_INSTALL_LIBDIR, according to CMake docs, is a relative path ([GNUInstallDirs doc](https://cmake.org/cmake/help/v3.22/module/GNUInstallDirs.html)):

> CMAKE_INSTALL_\<dir\>

> Destination for files of a given type. This value may be passed to the DESTINATION options of install() commands for the 
> corresponding file type. It should typically be a path relative to the installation prefix so that it can be converted to an 
> absolute path in a relocatable way (see CMAKE_INSTALL_FULL_\<dir\>). However, an absolute path is also allowed.

jansol:
My bad, the part I was thinking of is about the destination parameter to `install()` and it is described right in the intro section of [the page I linked](https://cmake.org/cmake/help/latest/command/install.html#introduction):

[...]

PR #769, opened 2021-11-15
Fix installation with absolute paths in CMAKE_INSTALL_*DIR

jansol:
See discussion in #764. See also libexpat/libexpat#459 for a different approach.

This produces correct results when CMAKE_INSTALL_LIBDIR and CMAKE_INSTALL_INCLUDEDIR are absolute paths.


tomfinegan:
Closing this: obsoleted by https://github.com/google/draco/pull/775

Please let me know if I missed something, or if you're still running into issues similar to what you reported in https://github.com/google/draco/issues/764

jansol:
With that change the build is indeed fine, thanks!

PR #777, opened 2021-12-04
Revert "Add CIFuzz integration"

tomfinegan:
Reverts google/draco#774

This is breaking at HEAD. It looks like a configuration problem: if that can be addressed we can revisit the PR.


tomfinegan:
Landing now because revert.

PR #796, opened 2022-01-21
Add install target testing.

tomfinegan:
Add a script the builds and installs Draco in shared and static configurations,
and confirms that a simple test application can compile and link using the CMake
configuration provided by each Draco installation.


ondys (approved):

cmake/draco-config.cmake.template
```diff
@@ -1,4 +1,7 @@
 @PACKAGE_INIT@
 set_and_check(DRACO_INCLUDE_DIR "@CMAKE_INSTALL_FULL_INCLUDEDIR@")
-set_and_check(DRACO_LIB_DIR "@CMAKE_INSTALL_FULL_LIBDIR@")
-set_and_check(DRACO_LIBRARY "draco")
+set_and_check(DRACO_LIBRARY_DIR "@CMAKE_INSTALL_FULL_LIBDIR@")
+set(DRACO_NAMES draco.dll libdraco.dylib libdraco.so draco.lib libdraco.a)
+find_library(_DRACO_LIBRARY PATHS ${DRACO_LIBRARY_DIR} NAMES ${DRACO_NAMES})
```
JackBoosY, line 5: I must say that this method will not work with mutiple configuration, such as debug **and** release:
In debug mode, it may find the release library.

tomfinegan: Are you merely speculating, or do you have a reproducible issue?

JackBoosY: According to the [latest doc](https://cmake.org/cmake/help/latest/command/find_library.html), `PATHS` specify directories to search in addition to the default locations.
cmake also can find other places such as `CMAKE_PREFIX_PATH`, `CMAKE_LIBRARY_PATH` if `NO_DEFAULT_PATH` not specified.

The cmake official has clearly stated that using `find_library` in the configration files is a deprecated way, although `DRACO_NAMES` can be synchronized through the settings in _CMakeLists.txt_, but the real library name is still affected by other options such as [LIBRARY_OUTPUT_NAME](https://cmake.org/cmake/help/latest/prop_tgt/LIBRARY_OUTPUT_NAME.html). 
Using the export method is more secure and is officially recommended by cmake.

tomfinegan: @JackBoosY Please file an issue or submit a PR. Discussing this in a stale PR is a not productive. 

Draco _is_ using exports in its install rule. Perhaps the current status is not not what you'd prefer, but that's where a PR or an issue would be useful. This PR is from February.

PR #810, opened 2022-02-09
Allow use of local third party dependencies

tomfinegan:
Add configuration variables that allow users to use already present local files instead of
forcing the use of submodules:

- DRACO_EIGEN_PATH - path to Eigen
- DRACO_FILESYSTEM_PATH - path to gulrak/filesystem
- DRACO_GOOGLETEST_PATH - path to googletest
- DRACO_TINYGLTF_PATH - path to syoyo/tinygltf


tomfinegan:
This is included in the 1.5.1 snapshot.

PR #825, opened 2022-03-01
fix M_PI error on MSVC

xxzl0130:
Without the macro `_USE_MATH_DEFINES`, MSVC will emit a compile error on M_PI

### DIFF
--- a/src/draco/scene/light.cc
+++ b/src/draco/scene/light.cc
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+#ifdef _MSC_VER
+#define _USE_MATH_DEFINES
+#endif
+
 #include "draco/scene/light.h"
 
 #ifdef DRACO_TRANSCODER_SUPPORTED


tomfinegan:
There's an internal fix for this that will be pushed today or tomorrow-- the fix in this PR is insufficient. Compilation will continue to fail within the tests. In addition, use of _USE_MATH_DEFINES to avoid this problem remains vulnerable to breakage via use of precompiled headers (PCH/stdafx.h) in user code. In other words, this is, in a way, a race condition wrt when _USE_MATH_DEFINES is defined, and when the include guard in math.h is evaluated. When _USE_MATH_DEFINES is defined before the first time math.h 

I've approved the run of the workflow, and I am nearly certain will likely pass: I am still in the process of adding transcoder enabled builders to our test configuration, and those configurations are required to catch issues like the one you're attempting to address in this PR. Try building Draco with your patch when DRACO_TESTS and DRACO_TRANSCODER_SUPPORTED are both enabled.

Issue #834, opened 2022-04-06
CMake: install dll in bin folder & allow to not install static


SpaceIm:
I maintain draco recipe in conan-center (public repository of conan package manager), and we have a recurrent patch related to CMake installation of draco:

- dll is installed in `lib` folder, which is not a standard layout on Windows where dll should go into `bin` folder.

- Except for msvc, static & shared are built & installed together since 1.4.0. It would be nice if there was a way to disable installation of static when `BUILD_SHARED_LIBS ON` (packaging both flavors in the same build is not allowed in conan-center, and I think also in vcpkg).

Basically here is our patch for 1.5.2:

```patch
--- a/cmake/draco_install.cmake
+++ b/cmake/draco_install.cmake
@@ -84,12 +84,14 @@ macro(draco_setup_install_target)
   endif()
 
   if(MSVC)
-    install(TARGETS draco DESTINATION "${libs_path}")
+    install(TARGETS draco RUNTIME DESTINATION bin ARCHIVE DESTINATION lib LIBRARY DESTINATION lib)
   else()
+    if(NOT BUILD_SHARED_LIBS)
     install(TARGETS draco_static DESTINATION "${libs_path}")
+    endif()
 
     if(BUILD_SHARED_LIBS)
-      install(TARGETS draco_shared DESTINATION "${libs_path}")
+      install(TARGETS draco_shared RUNTIME DESTINATION bin ARCHIVE DESTINATION lib LIBRARY DESTINATION lib)
     endif()
   endif()
 
```

SpaceIm:
I can submit a PR.
What would you prefer regarding installation of static lib? A `DRACO_INSTALL_STATIC` which is available if `NOT MSVC and BUILD_SHARED_LIBS`, and defaulted to ON to keep current behavior by default?

tomfinegan:
There are a couple things going on here:

1) The lib _must_ be installed with the dll. The lib file is actually an import library. Trying to use draco without the import library when BUILD_SHARED_LIBS is enabled will not work. It's required at link time when building executables that use the draco dll. 

2) I agree that the dll (and lib, because of the above) should be installed to the bin dir on Windows side-by-side with the tools. 

Regarding 1,  the draco install rule will not be changed re the installation of the import lib: doing so would break usage of the DLL for users using installed draco to link it into their projects. We're not breaking that functionality in order to appease opinionated package managers.

I'm looking into the second issue. It likely needs a bit more work to enable proper installation behavior in mingw/msys environments.

[...]

PR #838, opened 2022-04-08
CMake: install draco.dll in bin folder

SpaceIm:
closes https://github.com/google/draco/issues/834

### DIFF
--- a/cmake/draco-config.cmake.template
+++ b/cmake/draco-config.cmake.template
@@ -1,11 +1,6 @@
 @PACKAGE_INIT@
 set_and_check(DRACO_INCLUDE_DIR "@CMAKE_INSTALL_FULL_INCLUDEDIR@")
 set_and_check(DRACO_LIBRARY_DIR "@CMAKE_INSTALL_FULL_LIBDIR@")
-set(DRACO_NAMES
-    draco.dll libdraco.dylib libdraco.so draco.lib libdraco.dll libdraco.a)
-find_library(_DRACO_LIBRARY PATHS ${DRACO_LIBRARY_DIR} NAMES ${DRACO_NAMES})
+find_library(_DRACO_LIBRARY PATHS ${DRACO_LIBRARY_DIR} NAMES draco)
 set_and_check(DRACO_LIBRARY ${_DRACO_LIBRARY})
-find_file(DRACO_LIBRARY_DLL
-          PATHS ${DRACO_LIBRARY_DIR}
-          NAMES draco.dll libdraco.dll)
 set(DRACO_FOUND YES)
--- a/cmake/draco_install.cmake
+++ b/cmake/draco_install.cmake
@@ -90,9 +90,7 @@ macro(draco_setup_install_target)
       ARCHIVE DESTINATION "${libs_path}"
       LIBRARY DESTINATION "${libs_path}")
   else()
-    if(NOT BUILD_SHARED_LIBS OR DRACO_INSTALL_STATIC)
-      install(TARGETS draco_static DESTINATION "${libs_path}")
-    endif()
+    install(TARGETS draco_static DESTINATION "${libs_path}")
 
     if(BUILD_SHARED_LIBS)
       install(
--- a/cmake/draco_install.cmake
+++ b/cmake/draco_install.cmake
@@ -24,7 +24,6 @@ include(GNUInstallDirs)
 # target is created.
 macro(draco_setup_install_target)
   set(bin_path "${CMAKE_INSTALL_FULL_BINDIR}")
-  set(data_path "${CMAKE_INSTALL_FULL_DATAROOTDIR}")
   set(includes_path "${CMAKE_INSTALL_FULL_INCLUDEDIR}")
   set(libs_path "${CMAKE_INSTALL_FULL_LIBDIR}")
   set(draco_lib_name "draco")
@@ -39,10 +38,10 @@ macro(draco_setup_install_target)
   configure_package_config_file(
     "${draco_root}/cmake/draco-config.cmake.template"
     "${draco_build}/draco-config.cmake"
-    INSTALL_DESTINATION "${data_path}/cmake")
+    INSTALL_DESTINATION "${libs_path}/cmake/draco")
   install(
     FILES "${draco_build}/draco-config.cmake"
-    DESTINATION "${data_path}/cmake")
+    DESTINATION "${libs_path}/cmake/draco")
 
   # CMake version: draco-version.cmake
   write_basic_package_version_file(
@@ -51,7 +50,7 @@ macro(draco_setup_install_target)
     COMPATIBILITY AnyNewerVersion)
   install(
       FILES "${draco_build}/draco-version.cmake"
-      DESTINATION "${data_path}/cmake")
+      DESTINATION "${libs_path}/cmake/draco")
 
   foreach(file ${draco_sources})
     if(file MATCHES "h$")
--- a/cmake/draco_install.cmake
+++ b/cmake/draco_install.cmake
@@ -91,7 +91,9 @@ macro(draco_setup_install_target)
       LIBRARY DESTINATION "${libs_path}"
     )
   else()
-    install(TARGETS draco_static DESTINATION "${libs_path}")
+    if(NOT BUILD_SHARED_LIBS OR DRACO_INSTALL_STATIC)
+      install(TARGETS draco_static DESTINATION "${libs_path}")
+    endif()
 
     if(BUILD_SHARED_LIBS)
       install(
--- a/cmake/draco_install.cmake
+++ b/cmake/draco_install.cmake
@@ -43,13 +43,13 @@ macro(draco_setup_install_target)
     FILES "${draco_build}/draco-config.cmake"
     DESTINATION "${libs_path}/cmake/draco")
 
-  # CMake version: draco-version.cmake
+  # CMake version: draco-config-version.cmake
   write_basic_package_version_file(
-    "${draco_build}/draco-version.cmake"
+    "${draco_build}/draco-config-version.cmake"
     VERSION ${DRACO_VERSION}
     COMPATIBILITY AnyNewerVersion)
   install(
-      FILES "${draco_build}/draco-version.cmake"
+      FILES "${draco_build}/draco-config-version.cmake"
       DESTINATION "${libs_path}/cmake/draco")
 
   foreach(file ${draco_sources})
--- a/cmake/draco_install.cmake
+++ b/cmake/draco_install.cmake
@@ -88,8 +88,7 @@ macro(draco_setup_install_target)
       TARGETS draco
       RUNTIME DESTINATION "${bin_path}"
       ARCHIVE DESTINATION "${libs_path}"
-      LIBRARY DESTINATION "${libs_path}"
-    )
+      LIBRARY DESTINATION "${libs_path}")
   else()
     if(NOT BUILD_SHARED_LIBS OR DRACO_INSTALL_STATIC)
       install(TARGETS draco_static DESTINATION "${libs_path}")
@@ -100,8 +99,7 @@ macro(draco_setup_install_target)
         TARGETS draco_shared
         RUNTIME DESTINATION "${bin_path}"
         ARCHIVE DESTINATION "${libs_path}"
-        LIBRARY DESTINATION "${libs_path}"
-      )
+        LIBRARY DESTINATION "${libs_path}")
     endif()
   endif()
 
--- a/cmake/draco_install.cmake
+++ b/cmake/draco_install.cmake
@@ -84,12 +84,22 @@ macro(draco_setup_install_target)
   endif()
 
   if(MSVC)
-    install(TARGETS draco DESTINATION "${libs_path}")
+    install(
+      TARGETS draco
+      RUNTIME DESTINATION "${bin_path}"
+      ARCHIVE DESTINATION "${libs_path}"
+      LIBRARY DESTINATION "${libs_path}"
+    )
   else()
     install(TARGETS draco_static DESTINATION "${libs_path}")
 
     if(BUILD_SHARED_LIBS)
-      install(TARGETS draco_shared DESTINATION "${libs_path}")
+      install(
+        TARGETS draco_shared
+        RUNTIME DESTINATION "${bin_path}"
+        ARCHIVE DESTINATION "${libs_path}"
+        LIBRARY DESTINATION "${libs_path}"
+      )
     endif()
   endif()
 
--- a/cmake/draco_options.cmake
+++ b/cmake/draco_options.cmake
@@ -89,6 +89,8 @@ macro(draco_set_default_options)
                "Build plugin library for Maya." VALUE OFF)
   draco_option(NAME DRACO_TRANSCODER_SUPPORTED HELPSTRING
                "Enable the Draco transcoder." VALUE OFF)
+  draco_option(NAME DRACO_INSTALL_STATIC HELPSTRING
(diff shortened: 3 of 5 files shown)


src/draco/tools/install_test/CMakeLists.txt
```diff
@@ -46,11 +46,3 @@ target_include_directories(install_check PUBLIC "${DRACO_INCLUDE_DIR}")
...
 
 install(TARGETS install_check DESTINATION "${CMAKE_INSTALL_FULL_BINDIR}")
-
-if(BUILD_SHARED_LIBS AND WIN32)
-  # Copy the Draco DLL into the bin dir for Windows: Windows does not really
-  # have a concept of rpath, but it does look in the current directory by
-  # default when a program tries to load a DLL.
-  install(FILES "${DRACO_LIBRARY_DLL}"
-          DESTINATION "${CMAKE_INSTALL_FULL_BINDIR}")
-endif()
```
SpaceIm, line 56: It took me some time to understand the purpose of this install_test, so yes moving dll or injecting its location in PATH is still required while running `install_check` executable.

I believe that all this logic, as well as `CMAKE_PREFIX_PATH` & `CMAKE_INSTALL_RPATH` (`CMAKE_INSTALL_RPATH_USE_LINK_PATH` is far better) should be moved in `test.py`.

tomfinegan:
> It took me some time to understand the purpose of this install_test, so yes moving dll or injecting its location in PATH is still required while running install_check executable.

I'm not sure why its purpose isn't obvious: we're making sure the installed dev package can function by building something that requires draco.

> I believe that all this logic, as well as CMAKE_PREFIX_PATH & CMAKE_INSTALL_RPATH (enabling CMAKE_INSTALL_RPATH_USE_LINK_PATH externally is far better here) should be moved in test.py.

Discussions about this here aren't that productive. If you believe your above statement to be correct, please update your PR and we'll see how the CI run goes. 

> BUILD_SHARED_LIBS doesn't make sense here in the CMake paradigm. It is supposed to control shared/static of current libraries, not external dependencies.

I have no idea what this statement means. BUILD_SHARED_LIBS is _exactly_ how you control your build configuration per CMake documentation. Please elaborate.

tomfinegan:
> Exception: install_check run failed!
> exit_code: 3221225781

I believe the above exit code means the DLL not found at runtime.

SpaceIm:
> > BUILD_SHARED_LIBS doesn't make sense here in the CMake paradigm. It is supposed to control shared/static of current libraries, not external dependencies.
> 
> I have no idea what this statement means. BUILD_SHARED_LIBS is _exactly_ how you control your build configuration per CMake documentation. Please elaborate.
> 
> Edit to add:
> 
> From [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) (currently tagged as latest, but actually v3.23.0):
> 
> > If present and true, this will cause all libraries to be built shared unless the library was explicitly added as a static library.
> > This variable is often added to projects as an [option()](https://cmake.org/cmake/help/latest/command/option.html#command:option) so that each user of a project can decide if they want to build the project using shared or static libraries.

`BUILD_SHARED_LIBS` controls the libraries defined in `add_library()` of current CMakeLists, not external libraries already built and discovered by `find_package()`. This is why I say it's odd to rely on `BUILD_SHARED_LIBS` a CMakeLists where the only target is an executable. A CMakeLists is supposed to be agnostic regarding static/shared status of external dependencies, it's controlled externally (ie command line or Preset) by providing the path of the proper flavor with `CMAKE_PREFIX_PATH`.

tomfinegan:
> > > BUILD_SHARED_LIBS doesn't make sense here in the CMake paradigm. It is supposed to control shared/static of current libraries, not external dependencies.
> > 
> > 
> > I have no idea what this statement means. BUILD_SHARED_LIBS is _exactly_ how you control your build configuration per CMake documentation. Please elaborate.
> > Edit to add:
> > From [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) (currently tagged as latest, but actually v3.23.0):
> > > If present and true, this will cause all libraries to be built shared unless the library was explicitly added as a static library.
> > > This variable is often added to projects as an [option()](https://cmake.org/cmake/help/latest/command/option.html#command:option) so that each user of a project can decide if they want to build the project using shared or static libraries.
> 
> `BUILD_SHARED_LIBS` controls the libraries defined in `add_library()` of current CMakeLists, not external libraries already built and discovered by `find_package()`. This is why I say it's odd to rely on `BUILD_SHARED_LIBS` in a CMakeLists where the only target is an executable. A CMakeLists is supposed to be agnostic regarding static/shared status of external dependencies, it's controlled externally (ie command line or Preset) by providing the path of the proper flavor with `CMAKE_PREFIX_PATH`.

When the install test runs:

1. test.py builds and installs draco in static and shared configurations
    - Each config installs the result to its own custom prefix.
2.  test.py  builds and installs the test program in shared and static configurations.  
    - Each config installs the result to its own custom prefix.
3. test.py runs install_test in each configuration.

I suppose it's a bit hacky to abuse BUILD_SHARED_LIBS in the context of the install test program CMake build, but in the context of the test and what is being tested it makes perfect sense to me. I'm not sure why you're conflating the usage of BUILD_SHARED_LIBS in the test script with normal CMake usage or practices. Adding another variable is silly when BUILD_SHARED_LIBS can be reused in the test. As you say: it does not influence an executable only CMake build.

SpaceIm:
`draco-config.cmake` is not installed in a standard location. See https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure

```
<prefix>/                                                       (W)
<prefix>/(cmake|CMake)/                                         (W)
<prefix>/<name>*/                                               (W)
<prefix>/<name>*/(cmake|CMake)/                                 (W)
<prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/                 (U)
<prefix>/(lib/<arch>|lib*|share)/<name>*/                       (U)
<prefix>/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/         (U)
<prefix>/<name>*/(lib/<arch>|lib*|share)/cmake/<name>*/         (W/U)
<prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/               (W/U)
<prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/ (W/U)
```

I'll change the install destination.

tomfinegan:
> `draco-config.cmake` is not installed in a standard location. See https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure
> 
> ```
> <prefix>/                                                       (W)
> <prefix>/(cmake|CMake)/                                         (W)
> <prefix>/<name>*/                                               (W)
> <prefix>/<name>*/(cmake|CMake)/                                 (W)
> <prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/                 (U)
> <prefix>/(lib/<arch>|lib*|share)/<name>*/                       (U)
> <prefix>/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/         (U)
> <prefix>/<name>*/(lib/<arch>|lib*|share)/cmake/<name>*/         (W/U)
> <prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/               (W/U)
> <prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/ (W/U)
> ```
> 
> Currently it's `<prefix>/shared/cmake`.
> 
> I'll change the install destination (the most common destination is `${CMAKE_INSTALL_LIBDIR}/cmake/draco`).

I'm uncertain how you're reaching that conclusion for `$prefix/share/cmake`. That's where it was going prior to this PR, at least.

SpaceIm:
> I'm uncertain how you're reaching that conclusion for $prefix/share/cmake. That's where it was going prior to this PR, at least.

Yes, but the test was hiding an issue since `CMAKE_PREFIX_PATH` was set with the exact location of the config file, while common usage is to put the root install dir in `CMAKE_PREFIX_PATH`.

SpaceIm:
I also see that config version file name is wrong, it should be `draco-config-version.cmake`. See https://cmake.org/cmake/help/latest/command/find_package.html#search-modes

I'll fix this also.

tomfinegan:
> I also see that config version file name is wrong, it should be `draco-config-version.cmake`. See https://cmake.org/cmake/help/latest/command/find_package.html#search-modes
> 
> I'll fix this also.

Agreed on both counts-- I'm going to fix those out of band. I don't want to mix obvious bug fixes in with this PR that's adding functionality.

SpaceIm:
```
RUN command: /Users/runner/work/draco/draco/src/draco/tools/install_test/_test_install_shared/bin/install_check
COMMAND output:
dyld: Library not loaded: @rpath/libdraco.4.dylib
  Referenced from: /Users/runner/work/draco/draco/src/draco/tools/install_test/_test_install_shared/bin/install_check
  Reason: image not found
```

I guess `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` doesn't work here because config file doesn't provide a proper imported target which can used in target_link_libraries().
I'll revert this part since it would require too much change (I have a branch for this but it would require to change CMakeLists so that it builds either static or shared, and you have moved away from this logic since 1.4.0 :s ).

tomfinegan:
> ```
> RUN command: /Users/runner/work/draco/draco/src/draco/tools/install_test/_test_install_shared/bin/install_check
> COMMAND output:
> dyld: Library not loaded: @rpath/libdraco.4.dylib
>   Referenced from: /Users/runner/work/draco/draco/src/draco/tools/install_test/_test_install_shared/bin/install_check
>   Reason: image not found
> ```
> 
> I guess `CMAKE_INSTALL_RPATH_USE_LINK_PATH=ON` doesn't work here because config file doesn't provide a proper imported target which can be used in target_link_libraries(). I'll revert this part since it would require too much change (I have a branch for this but it changes CMakeLists so that it builds either static or shared, and you have moved away from this logic since 1.4.0 :s ).

Sounds good. The ability to build shared+static targets simultaneously on macos/linux is not something I want to do drop without a really good reason.

SpaceIm:
A good reason would be to provide a unique `draco::draco` target in CMake config file, regardless of shared/static flavor. It's not something you can achieve when you build both static & shared. But it's out of scope of this PR ;)

tomfinegan:
> A good reason would be to provide a unique `draco::draco` target in CMake config file, regardless of shared/static flavor. It's not something you can achieve when you build both static & shared (when you build both, folks usually provide 1 target name for static and another for shared, but IMOO it's not a good practice, and here it would be even worse since msvc doesn't build both). But it's out of scope of this PR ;)

CMake isn't the only build system draco needs to support, and MSVC is not the only toolchain.

cmake/draco_install.cmake
```diff
@@ -84,12 +83,24 @@ macro(draco_setup_install_target)
...
-    install(TARGETS draco DESTINATION "${libs_path}")
+    install(
+      TARGETS draco
+      RUNTIME DESTINATION "${bin_path}"
+      ARCHIVE DESTINATION "${libs_path}"
+      LIBRARY DESTINATION "${libs_path}"
+    )
   else()
-    install(TARGETS draco_static DESTINATION "${libs_path}")
+    if(NOT BUILD_SHARED_LIBS OR DRACO_INSTALL_STATIC)
```
tomfinegan, line 93: Why does the static library install behavior need to change now that you have the DESTINATION flags for the shared lib case above and below? I think the DRACO_INSTALL_STATIC flag can be dropped from the PR.

cmake/draco_install.cmake
```diff
@@ -84,12 +83,24 @@ macro(draco_setup_install_target)
   endif()
 
   if(MSVC)
-    install(TARGETS draco DESTINATION "${libs_path}")
+    install(
+      TARGETS draco
+      RUNTIME DESTINATION "${bin_path}"
+      ARCHIVE DESTINATION "${libs_path}"
+      LIBRARY DESTINATION "${libs_path}"
+    )
```
tomfinegan, line 91: nit: closing peren on line with final arg.

cmake/draco_options.cmake
```diff
@@ -89,6 +89,8 @@ macro(draco_set_default_options)
                "Build plugin library for Maya." VALUE OFF)
   draco_option(NAME DRACO_TRANSCODER_SUPPORTED HELPSTRING
                "Enable the Draco transcoder." VALUE OFF)
+  draco_option(NAME DRACO_INSTALL_STATIC HELPSTRING
```
tomfinegan, line 92: As above: I don't think this variable is needed any longer.

cmake/draco_install.cmake
```diff
@@ -39,19 +38,19 @@ macro(draco_setup_install_target)
   configure_package_config_file(
     "${draco_root}/cmake/draco-config.cmake.template"
     "${draco_build}/draco-config.cmake"
-    INSTALL_DESTINATION "${data_path}/cmake")
+    INSTALL_DESTINATION "${libs_path}/cmake/draco")
```
tomfinegan, line 41: Please revert to data_path, or just cherry-pick the changes from #839 into your branch.

cmake/draco-config.cmake.template
```diff
@@ -1,11 +1,6 @@
 @PACKAGE_INIT@
 set_and_check(DRACO_INCLUDE_DIR "@CMAKE_INSTALL_FULL_INCLUDEDIR@")
 set_and_check(DRACO_LIBRARY_DIR "@CMAKE_INSTALL_FULL_LIBDIR@")
-set(DRACO_NAMES
-    draco.dll libdraco.dylib libdraco.so draco.lib libdraco.dll libdraco.a)
-find_library(_DRACO_LIBRARY PATHS ${DRACO_LIBRARY_DIR} NAMES ${DRACO_NAMES})
+find_library(_DRACO_LIBRARY PATHS ${DRACO_LIBRARY_DIR} NAMES draco)
```
tomfinegan, line 4: This is much cleaner now. Thanks!

tomfinegan (changes requested):

cmake/draco_install.cmake
SpaceIm: It's unrelated to dll in bin dir. Installing both static & shared is a no-go in conan package manager. With this option we can avoid to patch `draco` forever.

tomfinegan: I think you need to automate application of that patch downstream-- conan having a weird obsession with disallowing install of static libs does not warrant maintenance and testing of a configuration.

cmake/draco_install.cmake
SpaceIm: solved in merge of master: https://github.com/google/draco/pull/838/commits/85abbb3a22986c7ca7b54b4296f92e15ad877cf9

cmake/draco_install.cmake
SpaceIm: conan doesn't forbid installation of static, it forbids to install both static & shared in the same prefix.

tomfinegan: Right, and I think that's a conan problem. Draco shouldn't be responsible for maintaining and testing a configuration because of an arbitrary rule imposed by a downstream package manager.

cmake/draco_install.cmake
SpaceIm: fixed

cmake/draco_install.cmake
SpaceIm: reverted in https://github.com/google/draco/pull/838/commits/586fa33968b6e7259ab651de4be7df46f0338bd5

tomfinegan (approved):

PR #839, opened 2022-04-08
Correct the install location for CMake configuration files.

tomfinegan:
Also updates the install test to correctly set prefix and allow
CMake's built-in search functionality to work properly.


tomfinegan:
Addresses config file installation path issues identified in #838

ondys (approved):

PR #840, opened 2022-04-09
CMake: provide `draco::draco` imported target in cmake config file

SpaceIm:
This a refactoring of the CMake config file to follow best practices:
- A `draco-targets.cmake` is generated with `install(EXPORT ...)`, and included in `draco-config.cmake`, which is the standard way to create a modern and robust config file. The target `draco::draco` is available in the CMake config file. The installed package is also fully relocatable for free now (no more `CMAKE_INSTALL_FULL_*` substitution).
- For users consuming Draco with `add_subdirectory()` (vendoring), `draco::draco` alias target is also provided, so that everything is coherent.
- Use `cxx_std_11` compile features for libraries & executables to:
  -  not hardcode C++11 standard, so that users can build Draco with another C++ standard. By default CMake will use the default standard of the compiler -except if it is lower than 11-, but user can inject `CMAKE_CXX_STANDARD` externally to force another standard.
  - transitively propagate min C++ standard requirement to downstream targets. For example in install_test, `main.cpp` has no C++11 feature by itself, and I've removed cxx_std_11 here, but cxx_std_11 will be injected by `draco::draco` target because C++11 requirement comes from Draco library.
- `find-draco.cmake` is removed since it's not installed and the name is broken anyway (`find_package()` doesn't seach this pattern).
- CMakeLists of install_test is updated to use `draco::draco` imported target.


tomfinegan (approved):

cmake/draco_install.cmake
```diff
@@ -109,4 +89,36 @@ macro(draco_setup_install_target)
...
+
+  export(
+    EXPORT dracoExport
+    NAMESPACE draco::
+    FILE "${draco_build}/draco-targets.cmake")
+
+  install(
+    EXPORT dracoExport
+    NAMESPACE draco::
+    FILE draco-targets.cmake
```
Tradias, line 116: Now that the `draco-config.cmake.template` is essentially empty, this could also be named `draco-config.cmake` directly, thereby removing the whole `configure_package_config_file` step above and its associated parts.

SpaceIm: For current master branch yes.
In the meantime, it scales better to keep targets definition outside of `draco-config.cmake` (external dependencies may be added -requiring find_dependency()-, components or whatever logic).

Tradias: I know, just providing it as an option for the Draco maintainers. I am personally in favour of removing dead code. After all, by the time Draco needs custom draco-config logic the CMake the mechanism for achieving that could have changed already.

Tradias:
Tested successfully on Windows CMake 3.23 and Ubuntu CMake 3.22.

PR #921, opened 2022-08-31
Fix javascript example

### LINKED ISSUE #737: Unable to run the example application
I followed the steps in https://github.com/google/draco/tree/master/javascript/example and the one in the codelabs https://codelabs.developers.google.com/codelabs/draco-3d/index.html#5 but both of them result in the same error. 

```
DRACOLoader.js:7 Uncaught TypeError: Class constructor ic cannot be invoked without 'new'
    at new THREE.DRACOLoader (DRACOLoader.js:7)
    at webgl_loader_draco.html:47
THREE.DRACOLoader @ DRACOLoader.js:7
(anonymous) @ webgl_loader_draco.html:47
```

In `DRACOLoader.js`

```
/**
 * @author Don McCurdy / https://www.donmccurdy.com
 */

THREE.DRACOLoader = function ( manager ) {

	THREE.Loader.call( this, manager ); // <--- this line fails with the error "Uncaught TypeError: Class constructor ic cannot be invoked without 'new'"

	this.decoderPath = '';
	this.decoderConfig = {};
	this.decoderBinary = null;
	this.decoderPending = null;

	this.workerLimit = 4;
	this.workerPool = [];
	this.workerNextTaskID = 1;
	this.workerSourceURL = '';

	this.defaultAttributeIDs = {
		position: 'POSITION',
		normal: 'NORMAL',
		color: 'COLOR',
		uv: 'TEX_COORD'
	};
	this.defaultAttributeTypes = {
		position: 'Float32Array',
		normal: 'Float32Array',
		color: 'Float32Array',
		uv: 'Float32Array'
	};

};
```

Is anyone able to run the example applications within this repository?

jokester:
1. Follow upstream `THREE.Loader` to use ES class (this should fix https://github.com/google/draco/issues/737 )
2. set decoder path to load draco assets properly (if we follow javascript/README to set up a http server, assets in `REPO/javascript/name` will have URL path `/javascript/name`

### DIFF
--- a/javascript/example/DRACOLoader.js
+++ b/javascript/example/DRACOLoader.js
@@ -2,638 +2,632 @@
  * @author Don McCurdy / https://www.donmccurdy.com
  */
 
-THREE.DRACOLoader = function ( manager ) {
+THREE.DRACOLoader = class DRACOLoader extends THREE.Loader {
 
-	THREE.Loader.call( this, manager );
+    constructor(manager) {
+        super(manager);
+        this.decoderPath = '';
+        this.decoderConfig = {};
+        this.decoderBinary = null;
+        this.decoderPending = null;
 
-	this.decoderPath = '';
-	this.decoderConfig = {};
-	this.decoderBinary = null;
-	this.decoderPending = null;
+        this.workerLimit = 4;
+        this.workerPool = [];
+        this.workerNextTaskID = 1;
+        this.workerSourceURL = '';
 
-	this.workerLimit = 4;
-	this.workerPool = [];
-	this.workerNextTaskID = 1;
-	this.workerSourceURL = '';
+        this.defaultAttributeIDs = {
+            position: 'POSITION',
+            normal: 'NORMAL',
+            color: 'COLOR',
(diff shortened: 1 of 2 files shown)


jokester:
The super simple fix in https://github.com/google/draco/issues/737#issuecomment-922239479 should work too, but I think modern ES class makes it easier to read.

tomfinegan:
Thanks you for the PR!

Unfortunately, the present plan is to remove the example from our next release. The example will be replace with a link to the three.js docs and example.

tomfinegan:
> Thanks you for the PR!
> 
> Unfortunately, the present plan is to remove the example from our next release. The example will be replace with a link to the three.js docs and example.

Actually: change of plans. We will be updating the example, but I won't be able to get to this PR right away. Thanks for your patience!

tomfinegan:
@jokester When you have the time, can you rebase and push an updated patch? Thanks!

jokester:
@tomfinegan Sure. Just rebased my commits to latest `master`. The fix should still work.

Issue #935, opened 2022-10-28
One test fails


yurivict:
```
[ RUN      ] VectorDTest.TestSquaredDistance
/disk-samsung/freebsd-ports/archivers/draco/work/draco-1.5.4/src/draco/core/vector_d_test.cc:40: Failure
Expected equality of these values:
  squared_distance
    Which is: 41.58
  result
    Which is: 41.58
/disk-samsung/freebsd-ports/archivers/draco/work/draco-1.5.4/src/draco/core/vector_d_test.cc:40: Failure
Expected equality of these values:
  squared_distance
    Which is: 42.22
  result
    Which is: 42.22
[  FAILED  ] VectorDTest.TestSquaredDistance (0 ms)
```

Version: 1.5.4
clang-14
FreeBSD 13.1

tomfinegan:
That's strange. Those values are correct. Can you rerun, and/or maybe drop some logging in there? These issues are not reproducing elsewhere, and I don't have access to a FreeBSD system.

tomfinegan:
Actually, I can repro it with clang-14 on debian. Investigating.

PR #936, opened 2022-10-28
draco: Update CI configs.

tomfinegan:
- Rename the existing configs that use gcc-10: add gcc-10 to the names.
- Add clang-14 configs.


tomfinegan:
The new build configurations will not work until https://github.com/google/draco/issues/935 is fixed.

FrankGalligan (approved):
Thanks!

tomfinegan:
Disable GltfDecoderTest.TestLoadUnsupportedTexCoordAttributes to work around the flakiness we're tracking in https://github.com/google/draco/issues/938

PR #999, opened 2023-04-18
Added a note that we added support for binary STL files in this release.

rhulha:
Added a note that we added support for binary STL files in this release.


ondys:
the support for stl has been in at least since 1.5.3 so this one doesn't seem like the right place for this

PR #1001, opened 2023-05-05
Build as shared library or static library

topazus:
In some Linux distribution, like Fedora, openSUSE or Gentoo, shared libraries are more prefered. Static libraries are suggested to be avoided to be included in packaging. It may be better to explicitly to only build shared lib or static lib by using `BUILD_SHARED_LIBS` option.
Ref: 
https://en.opensuse.org/openSUSE:Shared_library_packaging_policy#Best_Practices
https://docs.fedoraproject.org/en-US/packaging-guidelines/#_shared_libraries
https://projects.gentoo.org/qa/policy-guide/installed-files.html#pg0302
https://github.com/google/re2/issues/258


topazus:
@ondys Could you see and review this PR?

PR #1021, opened 2023-09-06
DecodeFromTextBuffer

rvkennedy:
Implemented LoadTextBuffer, DecodeFromTextBuffer, and DecodeFromTextBufferToScene to add missing GLTF text format functionality (i.e. load from memory). This allows GLTF-format decoding without using the default file handling code, for example.
This functionality already exists for GLB, but for text-format it has not been implemented.


rvkennedy:
The CLA is complete.

ondys:
Can you add a test file to gltf_decoder_test.cc that would show that this is working? I couldn't find any gltf that would fail with the existing API.

Issue #1029, opened 2023-11-24
Add cmake option to enable/disable installation


ptc-tgamper:
We pull Draco into our Cmake based project by using Cmake's FetchContent(), which more or less adds the Draco checkout directory to our project via add_subdirectory(). That in turn causes Draco's install rules to be added. Which we prefer not to happen, as we only use Draco as a dependency, but don't want it to be installes along our targets. So, in step with other Cmake based projects, simply add a Cmake option, like DRACO_INSTALL, to enable/disable Draco's install rules.

I'll post a PR shortly.

vrabaud:
Maybe it is too intrusive and unconventional to add such an option?
If you use CMake 3.28, you should use EXCLUDE_FROM_ALL, cf https://cmake.org/cmake/help/v3.28/module/FetchContent.html#command:fetchcontent_declare
Otherwise, I see you can have a custom solution here: https://stackoverflow.com/questions/65527126/disable-install-for-fetchcontent

ptc-tgamper:
It's not only about FetchContent. There are other ways I can do something akin to add_subdirectory(draco). Moreover, we are already doing EXCLUDE_FROM_ALL by default for everything. The latter is only about not adding the install rules to the ALL target. They are still added to the draco target itself, and any target that depends on draco will invoke them as part of its own installation.
I came up with this simple solution because I have seen it in plenty other projects, e.g. https://github.com/CLIUtils/CLI11, https://github.com/glfw/glfw, https://github.com/syoyo/tinygltf, https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator
Having this simple option gives control, one can force the desired behaviour, without having to rely on other cmake shenanigans :)

PR #1030, opened 2023-11-24
Add cmake option to enable/disable installation Issue #1029

ptc-tgamper:
draco_options.cmake - add DRACO_INSTALL option;
draco_install.cmake - check DRACO_INSTALL before adding install rules

### DIFF
--- a/cmake/draco_install.cmake
+++ b/cmake/draco_install.cmake
@@ -23,99 +23,101 @@ include(GNUInstallDirs)
 # Sets up the draco install targets. Must be called after the static library
 # target is created.
 macro(draco_setup_install_target)
-  set(bin_path "${CMAKE_INSTALL_BINDIR}")
-  set(data_path "${CMAKE_INSTALL_DATAROOTDIR}")
-  set(includes_path "${CMAKE_INSTALL_INCLUDEDIR}")
-  set(libs_path "${CMAKE_INSTALL_LIBDIR}")
+  if(DRACO_INSTALL)
+    set(bin_path "${CMAKE_INSTALL_BINDIR}")
+    set(data_path "${CMAKE_INSTALL_DATAROOTDIR}")
+    set(includes_path "${CMAKE_INSTALL_INCLUDEDIR}")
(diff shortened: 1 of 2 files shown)


cmake/draco_install.cmake
```diff
@@ -23,99 +23,101 @@ include(GNUInstallDirs)
...
 macro(draco_setup_install_target)
-  set(bin_path "${CMAKE_INSTALL_BINDIR}")
-  set(data_path "${CMAKE_INSTALL_DATAROOTDIR}")
-  set(includes_path "${CMAKE_INSTALL_INCLUDEDIR}")
-  set(libs_path "${CMAKE_INSTALL_LIBDIR}")
-
-  foreach(file ${draco_sources})
-    if(file MATCHES "h$")
-      list(APPEND draco_api_includes ${file})
+  if (DRACO_INSTALL)
```
ptc-tgamper, line 26: Because this is a macro we cannot simpliy do an early return().

jzern: This looks like it should work. A couple of comments:

1. What is the reason for the change though? Were you running into an issue in the cmake invocation?
2. Change this line to `if(DRACO_INSTALL)` (no space after `if`) to match the rest of the code.

jzern: > This looks like it should work. A couple of comments:
> 
> 1. What is the reason for the change though? Were you running into an issue in the cmake invocation?

I missed the reference to #1029. The second point should still be addressed.

> 2. Change this line to `if(DRACO_INSTALL)` (no space after `if`) to match the rest of the code.

ptc-tgamper: The reason for the change is a project where we pull in draco via cmake FetchContent as a dependency. We found that we didn't have any control over the installation of draco files, they always got installed if we invoked cmake install, be it for a single target or the ALL target. But draco is only linked as a private dependency, we don't want to install it, we don't want its files ending up in the installation directory.

ondys (approved):

Issue #1031, opened 2023-11-24
Win32: Warning on wrong preprocessor syntax


ptc-tgamper:
When compiling on Win32 with Visual Studio 2019 the following warning is logged by the compiler:
`draco\src\draco\io\file_writer_utils.cc(41,28): warning C4067: unexpected tokens following preprocessor directive - expected a newline`

Above warning is legit as the preprocessor check
`if defined(_WIN32) && not defined(__MINGW32__)`
is malformed.

Will post a PR shortly.

PR #1032, opened 2023-11-24
Win32: Warning on wrong preprocessor syntax Issue #1031

ptc-tgamper:
file_writer_utils.cc - fix preprocessor check and wrong path check

### DIFF
--- a/src/draco/io/file_writer_utils.cc
+++ b/src/draco/io/file_writer_utils.cc
@@ -38,11 +38,10 @@ bool DirectoryExists(const std::string &path_arg) {
   struct stat path_stat;
   std::string path = path_arg;
 
-#if defined(_WIN32) && not defined(__MINGW32__)
+#if defined(_WIN32) && !defined(__MINGW32__)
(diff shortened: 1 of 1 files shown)


src/draco/io/file_writer_utils.cc
```diff
@@ -38,11 +38,10 @@ bool DirectoryExists(const std::string &path_arg) {
   struct stat path_stat;
   std::string path = path_arg;
 
-#if defined(_WIN32) && not defined(__MINGW32__)
+#if defined(_WIN32) && !defined(__MINGW32__)
   // Avoid a silly windows issue: stat() will fail on a drive letter missing the
-  // trailing slash.
-  if (path.size() > 0 && path[path.size()] != '\\' &&
```
ptc-tgamper, line 44: Acessing path[path.size()] is simply wrong, that is not the last character of the string, it's the null terminator.

ondys (approved):
Thanks for the PR. Looks good.

PR #1064, opened 2024-05-10
perf(core/bit_utils.h): optimize MostSignificantBit default impl

servusdei2018:
The fallback implementation for `MostSignificantBit` was optimized, replacing the while loop with a series of bitwise operations. Tests on my system indicate a 3-4x speedup.


servusdei2018:
And here's a reference benchmark comparing the original and modified implementations: https://godbolt.org/z/fE8vvfcv9

Issue #1082, opened 2024-09-18
custom GeometryAttributes for high dimensional point clouds results in error in THREE.DRACOLoader


patrikant:
Hi, I am trying to encode something similar, like [Draco does not decode "high dimensional" point clouds](https://github.com/google/draco/issues/1020). I think I managed to do it with some custom GeometryAttributes, however, I am facing some issues. My ply files look like this:
```ply
format ascii 1.0
element vertex 332185
property float x
property float y
property float z
property float scale_0
property float scale_1
property float scale_2
property uchar f_dc_0
property uchar f_dc_1
property uchar f_dc_2
property uchar opacity
property uchar rot_0
property uchar rot_1
property uchar rot_2
property uchar rot_3
end_header
```

I tried to add additional GeometryAttributes for scale and rotation. I can encode the file and also decode it locally. However, decompiling on the web with Javascript does not work.
Here I get an error:
`THREE.DRACOLoader: Decoding failed: Failed to decode point attributes.`
Something in `draco.Decoder()` does not fit here, the error comes from the line: 
`decodingStatus = decoder.DecodeBufferToPointCloud( decoderBuffer, dracoGeometry );`
Any ideas on how or where I could change the Draco decoder to add these GeometryAttributes`? or what could be my error?

zihanzheng-sjtu:
I have the same question, how or where I could change the Draco decoder to add these GeometryAttributes?

Issue #1020, opened 2023-09-06
Draco does not decode "high dimensional" point clouds


SvenKratz:
I'm trying to apply draco to a "[Gaussian Splatting](https://github.com/graphdeco-inria/gaussian-splatting)" point cloud, which has many more dimensions than just point locations, e.g., here are the header fields:

```
ply
format binary_little_endian 1.0
element vertex 1167703
property float x
property float y
property float z
property float nx
property float ny
property float nz
property float f_dc_0
property float f_dc_1
property float f_dc_2
property float f_rest_0
property float f_rest_1
property float f_rest_2
property float f_rest_3
property float f_rest_4
property float f_rest_5
property float f_rest_6
property float f_rest_7
property float f_rest_8
property float f_rest_9
property float f_rest_10
property float f_rest_11
property float f_rest_12
property float f_rest_13
property float f_rest_14
property float f_rest_15
property float f_rest_16
property float f_rest_17
property float f_rest_18
property float f_rest_19
property float f_rest_20
property float f_rest_21
property float f_rest_22
property float f_rest_23
property float f_rest_24
property float f_rest_25
property float f_rest_26
property float f_rest_27
property float f_rest_28
property float f_rest_29
property float f_rest_30
property float f_rest_31
property float f_rest_32
property float f_rest_33
property float f_rest_34
property float f_rest_35
property float f_rest_36
property float f_rest_37
property float f_rest_38
property float f_rest_39
property float f_rest_40
property float f_rest_41
property float f_rest_42
property float f_rest_43
property float f_rest_44
property float opacity
property float scale_0
property float scale_1
property float scale_2
property float rot_0
property float rot_1
property float rot_2
property float rot_3
end_header
```

Draco seems to truncate all fields after the first six, assuming a coords + normals representation. Are there any plans to make the approach more general to support "exotic" point clouds such as those used by Gaussian splatting?

ondys:
This is a limitation of the ply decoder that is used in Draco. The point cloud compression itself supports arbitrary number of dimensions. A solution would be to parse the data on your end and initiate the `draco::PointCloud` data structure manually using the C++ api.

PR #1097, opened 2025-03-07
add SPLAT decoder to be able to encode .splat files to .drc files

patrikant:
I am using SPLAT files in addition to ply files. Currently, this datatype is not supported by the Draco library. A few months ago, I raised an issue regarding this topic (https://github.com/google/draco/issues/1082), and there is another one concerning Gaussian splatting (https://github.com/google/draco/issues/1020).

I am aware that SPLAT files are not standardized yet. This version accepts SPLAT files with the following attributes:
```struct Gaussian {
	float position[3];
	float scale[3];
	uint8_t color[4];
	uint8_t rotation[4];
};```
I attempted to encode the scale and rotation attributes as `GeometryAttribute::GENERIC` or custom attributes. However, I encountered some issues with the draco web loader in viewing these attributes. The two "new" attributes could not be decoded correctly. As a result, I am encoding position and scale together, as well as color and rotation. The implementation works perfectly for me.

Some examples for the compression using CL 10 and QP 11:
SPLAT file with 33MB -> DRC file with 10MB
SPLAT file with 4.6MB -> DRC file with 1.4MB
SPLAT file with 10MB -> DRC file with 3MB

### DIFF
--- a/src/draco/io/splat_decoder.h
+++ b/src/draco/io/splat_decoder.h
@@ -1,0 +1,44 @@
+// Copyright 2025 Patrick Trollmann.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#ifndef DRACO_IO_SPLAT_DECODER_H_
+#define DRACO_IO_SPLAT_DECODER_H_
+
+#include <fstream>
+#include <vector>
+
+#include "draco/point_cloud/point_cloud.h"
+#include "draco/core/status.h"
+
+namespace draco {
+
+// Decodes a SPLAT file into draco::PointCloud.
+// The current implementation assumes the properties: POSITION, SCALE, COLOR, ROTATION.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by waveviewer** [comment]
Following this PR, I compressed a splat file from 187M to 41M. Could you provide the code for decoding the DRC back to splat? Thanks! @patrikant

**Comment by patrikant** [comment]
> Could you provide the code for decoding the DRC back to splat?
@waveviewer I do not need a local Draco decoder to decode DRC to Splat. So, I haven't customized these files yet, unfortunately. If it's necessary for this PR to be included in the next build, I can do it.
I am using the DRC files solely to view them in a WebGL viewer. I utilize a custom function from `DracoLoader.js`. Since they deprecated the function for using custom attributes to decode the geometry (`dracoLoader.load` is deprecated, so `dracoLoader.decodeGeometry` only uses the standard attributes), I don't know why this decision was made...
The decoded geometry is then utilized with a SplatLoader to be added to the scene.

File: src/draco/io/splat_decoder.h [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+
+#include <fstream>
+#include <vector>
+
+#include "draco/point_cloud/point_cloud.h"
+#include "draco/core/status.h"
+
+namespace draco {
+
+// Decodes a SPLAT file into draco::PointCloud.
```
ondys, line 26: is there any official SPLAT file specification / definition that we could link to here?

src/draco/io/splat_decoder.h
```diff
@@ -0,0 +1,44 @@
...
+// The current implementation assumes the properties: POSITION, SCALE, COLOR, ROTATION.
+// POSITION and SCALE are stored together as GeometryAttribute.POSITION. COLOR and ROATION
```
ondys, line 28: typo: ROTATION

PR #1106, opened 2025-06-19
Allows disabling executables build for 3rd party usage

Raphael-Martin-TS3D:
To improve draco consumption as 3rd dependency, allow disabling executables build.
Note: Goal is not to change default behavior.


FrankGalligan:
Looks good. Thanks!

FrankGalligan (approved):
Looks good. Thanks!

PR #1118, opened 2025-09-04
support ply texcoord

HaruoOgawa:


ondys:
while adding texcoord support to ply decoding would be a nice feature to have, I'm not sure if the code in this PR would actually work for most cases. It would certainly not work for texture coordinates encoded with our `PlyEncoder` which would be probably the main requirement.
The most common way to store texcoord in ply is by using an face element list property, something like:
```
element face numFaces
...
property list uint8 float texcoord
```

Is there any software that produces per-vertex texcoords with `texture_u` and `texture_v` properties as proposed in this PR?

igorvytyaz:
Closing this PR due to inactivity.

PR #1121, opened 2025-09-16
Fix #1058

### LINKED ISSUE #1058: ld: unknown option: --start-group
draco 1.5.7 fails to link on macOS when using an llvm.org release of clang (as opposed to an Apple release of clang from Xcode or its command line tools). The error is:

```
ld: unknown option: --start-group
```

The cause is this code:

https://github.com/google/draco/blob/7d58126d076bc3f5f9d8c114d1700b7311faecfe/cmake/draco_targets.cmake#L165-L179

Apple clang identifies itself with the CMake compiler ID `AppleClang` (assuming [CMP0025](https://cmake.org/cmake/help/latest/policy/CMP0025.html) is set to `NEW` which in my case it is) and llvm.org clang identifies itself as just `Clang`. This code, therefore, is checking if the compiler is llvm.org clang (or GCC) and if so it is adding the linker flags `--start-group` and `--end-group` which the macOS linker does not understand. You are presuming that using a non-Apple build of clang (or using GCC) means compiling on a non-Apple operating system, which is not a valid assumption.

stepeos:
There is a logical flaw in the cmake script. Apple uses the ld64 linker which does not support `--start-group` and `--end-group`. The assumption, that clang builds always support these linker options depending on the compiler is false, since the compiler is interchangeable for the same linker on many systems. This PR fixes this bug for windows and apple builds using GNU and clang compilers.

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -248,7 +248,7 @@ GEM
     rb-fsevent (0.11.2)
     rb-inotify (0.11.1)
       ffi (~> 1.0)
-    rexml (3.4.1)
+    rexml (3.4.2)
     rouge (3.30.0)
     rubyzip (2.4.1)
     safe_yaml (1.0.5)


stepeos:
i submitted CLA, please re-run CI

igorvytyaz (approved):

igorvytyaz:
Thank you for making this contribution to Draco, @stepeos!

PR #1167, opened 2026-04-02
fix: validate face vertex indices in sequential decoder to prevent OOB access

mohammadmseet-hue:
## Summary

The sequential mesh decoder reads face vertex indices from the Draco bitstream without validating that they fall within `[0, num_points)`. When a crafted Draco file contains out-of-range indices, attribute access via `GeometryAttribute::GetValue()` causes an **out-of-bounds heap read**.

## ASan Proof

```
==ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 12 at 0x507000000540 thread T0
    #0 in memcpy
    #1 in draco::DataBuffer::Read (data_buffer.h:48)
    #2 in draco::GeometryAttribute::GetValue (geometry_attribute.h:138)

0x507000000540 is located 1128 bytes after 72-byte region [0x507000000090,0x5070000000d8)
```

A 72-byte attribute buffer (6 points × 12 bytes) was accessed at index 100 (offset 1200) — **1128 bytes past the end**.

## Root Cause

`mesh_sequential_decoder.cc` lines 74-125: all four index decoding paths (`uint8`, `uint16`, `uint32` varint, `uint32` direct) store indices from the stream into `face[j]` without checking `val < num_points`. The compressed index path (`DecodeAndDecompressIndices`) has the same issue.

## Impact

Any application that decodes attacker-provided `.drc` files is affected. The OOB read can leak heap data. Combined with attribute write operations (deduplication, transforms), this could potentially become an OOB write.

## Fix

Added a validation pass after decoding connectivity that checks all face vertex indices are within `[0, num_points)`.

## Separate from PR #1166

This is a **different root cause** from the integer overflow in PR #1166:
- PR #1166: `num_faces * 3` overflow → wrong allocation size → heap overflow (WRITE)
- This PR: unvalidated face indices → OOB attribute access (READ)

PR #1166, opened 2026-04-02
fix: integer overflow in face/point count multiplications → heap overflow

mohammadmseet-hue:
## Summary

Multiple locations in the Unity plugin, JavaScript/WASM decoder wrapper, and Maya plugin compute buffer sizes using `num_faces * 3` or `num_points * N` with attacker-controlled values from the Draco bitstream, without checking for integer overflow. When the product wraps to a small value, a tiny buffer is allocated and subsequent loops write the full (unwrapped) count of entries, causing a **heap buffer overflow**.

## Severity

**Critical** — Heap buffer overflow from crafted Draco files. Affects:
- Unity games loading user-provided 3D models
- Web applications using the Draco WASM decoder (Three.js, Google Model Viewer)
- Maya plugin loading untrusted meshes

### 1. JavaScript/WASM Decoder (decoder_webidl_wrapper.cc:178)
```cpp
const uint32_t num_faces = m.num_faces();
if (num_faces * 3 * sizeof(T) != out_size) {  // Overflow breaks this check
```
On WASM32, `sizeof(T)=4`, so `357913942 * 3 * 4 = 4294967304` wraps to `8`. If `out_size=8`, the check passes and the loop writes ~1.4B entries into an 8-byte buffer.

### 2. Unity Plugin (draco_unity_plugin.cc:239, 335, 343, 358, 373, 393)
```cpp
int *const temp_indices = new int[m->num_faces() * 3];  // Overflow → tiny alloc
unity_mesh->position = new float[in_mesh->num_points() * 3];
unity_mesh->normal = new float[in_mesh->num_points() * 3];
unity_mesh->color = new float[in_mesh->num_points() * 4];
unity_mesh->texcoord = new float[in_mesh->num_points() * 2];
```

### 3. Maya Plugin (draco_maya_plugin.cc:23)
```cpp
out_mesh->faces = new int[num_faces * 3];  // Overflow → tiny alloc
```

## Trigger

`num_faces = 1431655766` (just over UINT32_MAX / 3):
- `1431655766 * 3 = 4294967298` → overflows 32-bit to `2`
- `new int[2]` allocates 8 bytes
- Loop writes `1431655766 * 3 = ~4.3 billion` entries → **heap overflow**

## Note

[...]

PR #1168, opened 2026-04-02
fix: integer overflow in CopyDataAsUint32 accessor.count * num_components

mohammadmseet-hue:
## Summary

In `CopyDataAsUint32()` (gltf_decoder.cc:170), the computation `accessor.count * num_components` uses `int` arithmetic. When a crafted glTF file has a large `accessor.count` with a multi-component type (VEC2/VEC3/VEC4), the product overflows `int32`, causing `output.resize()` to receive a wrong value. The loop then writes `accessor.count * num_components` entries into the undersized buffer → **heap buffer overflow**.

## ASan Proof

```
poc_gltf_overflow.cc:31: warning: integer overflow in expression of type 'int' results in '-2147483647'

==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000000020
WRITE of size 4 at 0x502000000020 thread T0
    #0 in main poc_gltf_overflow.cc:74

0x502000000020 is located 0 bytes after 16-byte region [0x502000000010,0x502000000020)
```

## Trigger

`accessor.count = 715827883`, `num_components = 3` (VEC3):
- `715827883 * 3 = 2147483649` → overflows `int32` to `-2147483647`
- `output.resize(-2147483647)` → undefined behavior / tiny allocation
- Loop writes 2147483649 entries → heap overflow

## Fix

Use `int64_t` for the multiplication and check for overflow before casting to `int`.

## Separate from PRs #1166 and #1167

- PR #1166: `num_faces * 3` overflow in Unity/WASM/Maya binding layer
- PR #1167: unvalidated face vertex indices in sequential decoder
- **This PR**: `accessor.count * num_components` overflow in glTF decoder

Three separate root causes, three separate code paths.

### DIFF
--- a/src/draco/io/gltf_decoder.cc
+++ b/src/draco/io/gltf_decoder.cc
@@ -167,7 +167,14 @@ StatusOr<std::vector<uint32_t>> CopyDataAsUint32(
       tinygltf::GetComponentSizeInBytes(accessor.componentType);
   const int num_components =
       TinyGltfUtils::GetNumComponentsForType(accessor.type);
-  const int num_elements = accessor.count * num_components;
+  // Check for integer overflow in accessor.count * num_components.
+  const int64_t num_elements_64 =
+      static_cast<int64_t>(accessor.count) * num_components;
+  if (num_elements_64 > std::numeric_limits<int>::max()) {
+    return Status(Status::DRACO_ERROR,
+                  "CopyDataAsUint32: accessor count overflow.");
+  }
+  const int num_elements = static_cast<int>(num_elements_64);
 
   std::vector<uint32_t> output;
   output.resize(num_elements);


mohammadmseet-hue:
## ASan Confirmation

Built with `-fsanitize=address`:

```
poc_gltf_overflow.cc:31:45: warning: integer overflow in expression of type 'int' results in '-2147483647' [-Woverflow]

==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000000020
WRITE of size 4 at 0x502000000020 thread T0
    #0 in main poc_gltf_overflow.cc:74

0x502000000020 is located 0 bytes after 16-byte region [0x502000000010,0x502000000020)
allocated by thread T0 here:
    #0 in operator new(unsigned long)
    #1 in std::vector<unsigned int>::resize
    #6 in main poc_gltf_overflow.cc:68

SUMMARY: AddressSanitizer: heap-buffer-overflow poc_gltf_overflow.cc:74 in main
```

GCC also warns at compile time: `integer overflow in expression of type 'int' results in '-2147483647'`, confirming the overflow in the exact computation pattern from `gltf_decoder.cc:170`.

PR #1176, opened 2026-04-15
Fix memory safety bugs in PLY, OBJ, and STL decoders

sharadboni:
## Summary

This PR fixes three memory safety vulnerabilities in Draco's mesh file decoders:

### 1. PLY heap-buffer-overflow in `ply_reader.cc`
In `ParseElementData`, the `num_entries` field of a list property is read from attacker-controlled input and used to compute `num_bytes_to_read` without validating against the remaining buffer size. A crafted PLY file with a large list count causes an out-of-bounds read past the input buffer via the `prop.data_.insert(... buffer->data_head() + num_bytes_to_read)` call.

**Fix:** Added bounds check that `num_entries >= 0` and `num_bytes_to_read <= buffer->remaining_size()` before the insert.

### 2. OBJ negative-index integer underflow in `obj_decoder.cc`
OBJ files support negative vertex indices (e.g., `f -1 -2 -3`) which are relative to the current vertex count. However, there was no validation that `|index| <= count`. A crafted file like `f -99999` with only a few vertices causes `num_positions_ + indices[0]` to underflow to a huge value, producing an out-of-range `AttributeValueIndex`.

**Fix:** Added validation that `-indices[i] <= num_positions_` (and similarly for tex coords and normals) for negative indices, returning an error status on violation.

### 3. STL unchecked Decode return values in `stl_decoder.cc`
In `DecodeFromBuffer`, all calls to `buffer->Decode()` had their boolean return values ignored. A truncated binary STL file causes reads past the end of the buffer without any error reporting.

**Fix:** Check return values of `Decode()` and `remaining_size()` before each buffer operation, returning descriptive `IO_ERROR` status on truncated input.

## Test plan
- [ ] Verify existing draco unit tests still pass
- [ ] Test with crafted PLY file containing oversized list count
- [ ] Test with OBJ file containing out-of-range negative indices (e.g., `f -99999`)
- [ ] Test with truncated binary STL file (< 84 bytes)

### DIFF
--- a/src/draco/io/obj_decoder.cc
+++ b/src/draco/io/obj_decoder.cc
@@ -447,7 +447,11 @@ bool ObjDecoder::ParseFace(Status *status) {
       for (int c = 0; c < 3; c++) {
         const PointIndex vert_id(3 * num_obj_faces_ + c);
         const int triangulated_index = Triangulate(t, c);
-        MapPointToVertexIndices(vert_id, indices[triangulated_index]);
+        if (!MapPointToVertexIndices(vert_id, indices[triangulated_index])) {
+          *status = Status(Status::DRACO_ERROR,
+                           "OBJ: negative index out of range");
+          return true;
+        }
         // Save info about new edges that will allow us to reconstruct polygons.
         if (added_edge_att_id_ >= 0) {
           const AttributeValueIndex avi(IsNewEdge(nt, t, c));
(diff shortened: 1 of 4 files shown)


sharadboni:
@jzern Could you review this security fix? It addresses a heap-buffer-overflow in the PLY parser, an integer underflow in the OBJ parser, and unchecked return values in the STL decoder.

igorvytyaz:
Closing this PR as merged #1205 and #1203 address these issues.

PR #1191, opened 2026-04-17
Validate vertex indices in GetParallelogramEntries

TristanInSec:
## Summary

- Add bounds checks for vertex indices returned by `CornerTable::Vertex()` before using them as indices into `vertex_to_data_map`
- Return sentinel values (-1) when vertex indices are invalid or out of range
- Prevents out-of-bounds access when the corner table contains degenerate entries

## Test plan

- [x] Verified the fix handles `kInvalidVertexIndex` (UINT32_MAX) correctly
- [x] Verified the fix handles indices >= `vertex_to_data_map.size()` correctly
- [x] Existing callers already handle -1 return values (skip prediction)

PR #1192, opened 2026-04-17
Add missing bounds check in RAnsDecoder::read_init

TristanInSec:
## Summary

- Add `offset < 4` guard to the `x == 3` branch in `RAnsDecoder::read_init()`
- The other branches (`x == 1`, `x == 2`) already validate their respective minimum offsets before reading
- Prevents reading before the start of the buffer when offset is less than 4

## Test plan

- [x] Verified the fix rejects small offsets in the x==3 path
- [x] Confirmed existing branches (x==0, x==1, x==2) are unchanged
- [x] Matches the guard pattern used by the legacy `ans_read_init` C function


TristanInSec:
Hi, friendly ping on this security fix (see also #1191). Happy to make any changes needed. Thanks!

PR #1196, opened 2026-04-24
use FindPython instead of include(FindPythonInterp)

MerinoSheep:
FindPythonInterp has been deprecated since Cmake 3.12 - [CMP0148](https://cmake.org/cmake/help/latest/policy/CMP0148.html), replaced with [FindPython](https://cmake.org/cmake/help/latest/module/FindPython.html#module:FindPython)

No version bump needed since we are on 3.12 already


igorvytyaz (approved):

igorvytyaz:
Thank you for making this contribution to Draco, @MerinoSheep!

PR #1197, opened 2026-04-27
security: add bounds checks to decoder prediction schemes and point count

sharadboni:
## Summary

This PR fixes five heap-buffer-overflow and negative-integer vulnerabilities
in the Draco 3D decoder discovered through security analysis.

### Vulnerabilities 1–4: Ignored `size` parameter in mesh prediction scheme decoders

The `ComputeOriginalValues` method in four mesh prediction scheme decoders
(`MeshPredictionSchemeParallelogramDecoder`,
`MeshPredictionSchemeConstrainedMultiParallelogramDecoder`,
`MeshPredictionSchemeMultiParallelogramDecoder`,
`MeshPredictionSchemeGeometricNormalDecoder`) had the `size` parameter
commented out (`int /* size */`) and used `corner_map_size` from mesh data
as the loop bound instead.

When `corner_map_size > size / num_components`, the loop writes past the end
of `out_data`, causing a heap buffer overflow. The corner map and attribute
arrays are both populated from compressed data, so an attacker can craft a
DRC file where they diverge.

**Fix:** Restore the `size` parameter and add a guard:
```cpp
if (corner_map_size * num_components > size) {
  return false;
}
```

### Vulnerability 5: Negative point count in PointCloudSequentialDecoder

`DecodeGeometryData` reads `num_points` as `int32_t` from the bitstream
without checking for negative values. A negative value implicitly converts
to a very large `uint32_t` when passed to `set_num_points`, causing
downstream heap overflow.

**Fix:** Reject negative point counts:
```cpp
if (num_points < 0) {
  return false;
}
```

## Testing

All 25 existing `.drc` test files in `testdata/` decode correctly with the
fixed decoder under AddressSanitizer.

### DIFF
--- a/src/draco/compression/attributes/prediction_schemes/mesh_prediction_scheme_constrained_multi_parallelogram_decoder.h
+++ b/src/draco/compression/attributes/prediction_schemes/mesh_prediction_scheme_constrained_multi_parallelogram_decoder.h
@@ -82,7 +82,7 @@ template <typename DataTypeT, class TransformT, class MeshDataT>
 bool MeshPredictionSchemeConstrainedMultiParallelogramDecoder<
     DataTypeT, TransformT, MeshDataT>::
     ComputeOriginalValues(const CorrType *in_corr, DataTypeT *out_data,
-                          int /* size */, int num_components,
+                          int size, int num_components,
                           const PointIndex * /* entry_to_point_id_map */) {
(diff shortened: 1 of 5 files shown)


igorvytyaz (approved):

igorvytyaz:
Thank you for making this contribution to Draco, @sharadboni!

PR #1198, opened 2026-04-28
Fix missing bounds checks on decoded face indices in sequential mesh decoder

sharadboni:
## Summary

Two missing bounds checks in `MeshSequentialDecoder::DecodeConnectivity` allow
attacker-controlled face indices from a malformed `.drc` file to exceed
`num_points`, resulting in a heap out-of-bounds read (verified crash under
AddressSanitizer; silent memory corruption on release builds).

### Fix 1 — raw index paths (`mesh_sequential_decoder.cc` ~line 81)

The uncompressed connectivity paths (8-bit, 16-bit, varint, and 32-bit
branches) decoded each face index `val` directly from the bitstream without
checking `val < num_points`.  A check is now added immediately after each
successful decode:

```cpp
if (val >= num_points) {
  return false;
}
```

### Fix 2 — entropy-compressed path (`mesh_sequential_decoder.cc` ~line 170)

`DecodeAndDecompressIndices` reconstructed absolute indices via delta-decode
but never validated the result against `num_points`.  The method now accepts
`num_points` as a second `uint32_t` parameter and rejects any index outside
`[0, num_points)`:

```cpp
if (index_value < 0 ||
    static_cast<uint32_t>(index_value) >= num_points) {
  return false;
}
```

The declaration in `mesh_sequential_decoder.h` is updated to match.

## Affected files

- `src/draco/compression/mesh/mesh_sequential_decoder.cc`
- `src/draco/compression/mesh/mesh_sequential_decoder.h`

### DIFF
--- a/src/draco/compression/mesh/mesh_sequential_decoder.cc
+++ b/src/draco/compression/mesh/mesh_sequential_decoder.cc
@@ -65,7 +65,7 @@ bool MeshSequentialDecoder::DecodeConnectivity() {
     return false;
   }
   if (connectivity_method == 0) {
-    if (!DecodeAndDecompressIndices(num_faces)) {
+    if (!DecodeAndDecompressIndices(num_faces, num_points)) {
       return false;
     }
   } else {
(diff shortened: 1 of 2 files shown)


igorvytyaz (approved):

igorvytyaz:
Thank you for making this contribution to Draco, @sharadboni!

PR #1203, opened 2026-06-08
Fix heap OOB read in STL decoder from missing buffer bounds checks

gn00295120:
## Summary

`StlDecoder::DecodeFromBuffer` reads an 80-byte header, a 4-byte face count, and 50 bytes per face from the input buffer without checking that the buffer is large enough. A truncated or malicious STL file triggers a heap-buffer-overflow read.

## Changes

- Reject inputs smaller than 84 bytes (80-byte header + 4-byte count) before any read
- Check the `Decode()` return value for the face count (was silently ignored)
- Validate `face_count` against `remaining_size() / 50` before entering the parse loop
- Check per-face `Decode()` return values inside the loop

## Reproduction

```bash
# Build with ASAN
cmake --build build

# 1-byte STL file triggers heap-buffer-overflow READ of size 4
printf 'x' > crash.stl
./build/draco_encoder -i crash.stl
```

ASAN trace (before fix):
```
==PID==ERROR: AddressSanitizer: heap-buffer-overflow on address ...
READ of size 4 at ... thread T0
    #0 in draco::DecoderBuffer::Decode(void*, unsigned long) decoder_buffer.h:80
    #1 in draco::StlDecoder::DecodeFromBuffer stl_decoder.cc:46
```

### DIFF
--- a/src/draco/io/stl_decoder.cc
+++ b/src/draco/io/stl_decoder.cc
@@ -37,13 +37,25 @@ StatusOr<std::unique_ptr<Mesh>> StlDecoder::DecodeFromFile(
 
 StatusOr<std::unique_ptr<Mesh>> StlDecoder::DecodeFromBuffer(
     DecoderBuffer *buffer) {
+  // Binary STL requires at least an 80-byte header + 4-byte face count.
+  if (buffer->remaining_size() < 84) {
+    return Status(Status::IO_ERROR, "STL file too small.");
+  }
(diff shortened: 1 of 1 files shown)


igorvytyaz (approved):

gn00295120:
Thanks

igorvytyaz:
Thank you for making this contribution to Draco, @gn00295120!

PR #1204, opened 2026-06-08
Fix invalid pointer dereference in OBJ decoder for faces without vertices

gn00295120:
## Summary

`ObjDecoder` crashes when an OBJ file contains face definitions (`f 1 2 3`) but no preceding vertex positions (`v ...`). The position attribute ID stays at its initial value of -1, and `PointCloud::attribute(-1)` dereferences out of bounds.

Additionally, negative OBJ vertex indices (e.g., `f -1 -2 -3`) are resolved against `num_positions_` without checking for underflow.

## Changes

- After the counting pass, reject files that have faces but zero vertex positions (early error instead of UB)
- Guard `MapPointToVertexIndices` against `pos_att_id_ < 0`
- Validate that resolved negative indices are non-negative before use

## Reproduction

```bash
# Build with ASAN
cmake --build build

# 8-byte OBJ file triggers SEGV
printf 'f 1 2 3\n' > crash.obj
./build/draco_encoder -i crash.obj
```

ASAN trace (before fix):
```
==PID==ERROR: AddressSanitizer: SEGV on unknown address 0xfffffffffffffff8
    #0 in draco::PointCloud::attribute(int) point_cloud.h:89
    #1 in draco::ObjDecoder::MapPointToVertexIndices obj_decoder.cc:643
    #2 in draco::ObjDecoder::ParseFace obj_decoder.cc:450
```

### DIFF
--- a/src/draco/io/obj_decoder.cc
+++ b/src/draco/io/obj_decoder.cc
@@ -107,6 +107,10 @@ Status ObjDecoder::DecodeInternal() {
   }
 
   bool use_identity_mapping = false;
+  if (num_obj_faces_ > 0 && num_positions_ == 0) {
+    return Status(Status::DRACO_ERROR,
+                  "Faces reference vertex positions but none are defined.");
+  }
   if (num_obj_faces_ == 0) {
     // Mesh has no faces. In this case we try to read the geometry as a point
     // cloud where every attribute entry is a point.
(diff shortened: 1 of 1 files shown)


igorvytyaz (approved):

igorvytyaz (approved):

gn00295120:
nice , thanks!

igorvytyaz:
Thank you!

PR #1215, opened 2026-08-11
Reject mismatched point counts in legacy KD-tree streams

### LINKED ISSUE #1207: OSS-Fuzz issue 525708569
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=6166211298394112 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=525708569 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

carrerasdarren-cell:
## Summary

- reuse the bounded point decoder for legacy (pre-2.3) integer KD-tree streams
- reject disagreement between the outer attribute point count and the nested KD-tree count
- add a regression test using the public OSS-Fuzz reproducer

## Impact

The reproducer declares 168 points in the outer legacy attribute stream but 3,892,261,426 points in the nested KD-tree stream. The legacy path called the unbounded decoder overload, while its output iterator silently ignores writes beyond the allocated 168 entries. The decoder therefore continued processing the attacker-controlled nested count.

On an ASan build from `4723893`, the input was accepted after 10.04 seconds. With this change, it is rejected before entropy decoding in 0.04 seconds.

## Testing

- `PointCloudKdTreeEncodingTest.*`: 10/10 tests pass under ASan
- exact OSS-Fuzz reproducer is rejected
- 14 valid bitstream-2.2 KD-tree files generated by Draco 1.2.5 still decode, covering compression settings 1-10 and position-plus-color attributes
- `clang-format --dry-run --Werror`
- `git diff --check`

Fixes #1207


igorvytyaz (approved):

igorvytyaz:
Thank you for making this contribution to Draco, @carrerasdarren-cell!

PR #1225, opened 2026-09-10
decoder: avoid signed overflow in texture prediction

Alb3e3:
## Summary

The minimized OSS-Fuzz testcase for issue 553263223 reaches the portable texture-coordinate predictor with untrusted values that overflow signed `int64_t` during scaled UV addition. UBSan reports the overflow in `VectorD::operator+`.

Perform the intermediate vector arithmetic in `uint64_t` before converting back to the signed vector. Valid bitstreams retain their results, while malformed geometry no longer invokes signed-overflow undefined behavior during decoding.

Validation: rebuilt Draco with AddressSanitizer and UndefinedBehaviorSanitizer and replayed the minimized point-cloud and mesh decoder harnesses without findings.

This is a Google Patch Rewards OSS-Fuzz Tier 2 patch. AI assistance was used and is disclosed.


igorvytyaz (approved):

igorvytyaz:
Thank you for making this contribution to Draco, @Alb3e3!

